PR #15, opened 2015-05-21
Apply the fix for https://issues.apache.org/jira/browse/TOMEE-1590 in…

thusithathilina:
Please review and merge
This is to fix the https://issues.apache.org/jira/browse/TOMEE-1590 in TomEE 1.7.x branch


tomee/tomee-common/src/main/java/org/apache/tomee/common/WsFactory.java
```diff
@@ -17,6 +17,7 @@
  */
 package org.apache.tomee.common;
 
+import javafx.print.Collation;
```
danielsoro: JavaFX?

PR #28, opened 2015-11-27
walkmod cfg

rpau:
I have upgraded walkmod to allow executions without code transformations (just writing files). You need to donwload the 2.1.0 version. To analyze the results, you just need to execute:

```
walkmod apply
```

If you want to execute walkmod and you have not changes into the plugins you are using, is faster to execute:

```
walkmod apply --offline
```

If it is ok for you, I can proceed to:

1) Create the checkstyle plugin to automatically correct checkstyle problems.
2) Upgrade the pom.xml file to launch walkmod from the build process.


rmannibucau:
added the file, thks for the PR!

PR #32, opened 2016-05-27
Avoid NPE in RemoteTomEEContainer.start()

robertpanzer:
If the Arquillian configuration contains errors or some other problems occur during configuration processing a NullPointerException occurs in the catch block in `RemoteTomEEConfiguration.start()` because the container might not be initialized yet.

In this case the user will not get any feedback about the real error, because logging is done afterwards.

This PR simply only invokes `container.destroy()` if the container is not null.


rmannibucau:
think it can't happen anymore right?

robertpanzer:
It looks like it wouldn't happen on master.
But imo it could still happen on the 1.7.x branch.

radcortez:
Merged here: https://github.com/tomitribe/tomee/commit/f38c5098b88aeae9b010f2879c4d7f22add237f7

PR #34, opened 2016-07-02
Fix unused imports

ivanjunckes:
There were several imports that were not being used.


rmannibucau:
Applied, thanks Ivan!

PR #35, opened 2016-07-03
unused imports

tbookless:
Removed unused imports and cleaned up other warnings.


rmannibucau:
this class is a jaxb one so needs to be public, the attribute can stay protected for consistent. However note that this is part of openejb-jee which is generated when we upgrade the specs and this diff can be resetted anytime without any notice.

PR #36, opened 2016-07-26
Ensure deployment record is removed from deployments.xml even if dest…

zeppelinux:
…royApplication throws an exception, more info here: http://tomee-openejb.979440.n4.nabble.com/tomee-undeploy-is-not-clean-after-long-timeout-since-the-last-deployment-td4679481.html


r-tt:
Applied, thanks a lot!

radcortez:
Merged here: https://github.com/apache/tomee/commit/d94404e00983aa1383316c5702c5b5f34847751b

PR #39, opened 2016-09-13
Do not leak input stream in HttpResponseImpl

SvetlinZarev:
Bug: https://issues.apache.org/jira/browse/TOMEE-1932


SvetlinZarev:
I'll close this one, because this might break the connection pooling. I'll do some more research on the topic and come back to the issue if necessary.

PR #42, opened 2016-09-14
Use proper string equality checks

SvetlinZarev:
Use equals() instead of == / !=


SvetlinZarev:
This should have not been a PR.

PR #44, opened 2016-09-14
Do not leak resource streams

SvetlinZarev:
Do not leak streams from files or classloader.getResourceAsStream()

Bug: https://issues.apache.org/jira/browse/TOMEE-1935


SvetlinZarev:
Separated the changes to several commits to make it more readable and easy to review

PR #45, opened 2016-09-15
Fix - throw exception in finally block

kal0ian:


radcortez:
Merged here: https://github.com/apache/tomee/commit/2ee7a5b1545f54f6302bfd130dbefaf4dd4d8c6c

PR #52, opened 2016-11-04
[TOMEE-1974] add authentication header 

exabrial:
[wip] add authentication header


server/openejb-client/src/main/java/org/apache/openejb/client/JNDIContext.java
```diff
@@ -244,14 +243,16 @@ public Context getInitialContext(final Hashtable environment) throws NamingExcep
...
-                    " invalid: " +
-                    providerUrl +
-                    " - " +
-                    e.getMessage()).initCause(e);
+            throw (ConfigurationException) new ConfigurationException(
+                    "Property value for " + Context.PROVIDER_URL + " invalid: " + providerUrl + " - " + e.getMessage())
+                            .initCause(e);
         }
         this.server = new ServerMetaData(location);
+        String securityPrincipal = (String) env.get(Context.SECURITY_PRINCIPAL);
```
radcortez: Use different vendor properties for the Authentication header credentials, to allow the usage of different credentials for the JNDI call, or to omit the JNDI authentication.

server/openejb-client/src/main/java/org/apache/openejb/client/JNDIContext.java
```diff
@@ -244,14 +243,16 @@ public Context getInitialContext(final Hashtable environment) throws NamingExcep
...
+        String securityCredentials = (String) env.get(Context.SECURITY_CREDENTIALS);
+        if (securityPrincipal != null) {
```
radcortez: Code as been reordered and reformatted, but it seems that everything could stay as original and only add this piece.

server/openejb-ejbd/src/main/java/org/apache/openejb/server/ejbd/EjbRequestHandler.java
```diff
@@ -103,11 +103,17 @@ public Response processRequest(final ObjectInputStream in, final ProtocolMetaDat
...
         boolean failed = false;
         final CallContext call;
+        Object oldClientIdentity = null;
+
 
         try {
             try {
                 final Object clientIdentity = req.getClientIdentity();
                 if (clientIdentity != null) {//noinspection unchecked
+                    if (securityService.getCallerPrincipal() != null) {
```
radcortez: Not exactly sure why this is needed. When debugging securityService.disassociate() comes as null. Can you clarify why it was added?

radcortez:
@exabrial Could you clarify the reasoning behind this feature, plus documentation on how to use it? Thank you.

radcortez:
@jgallimore @exabrial I've enhanced the PR with:
- Separate properties to send in the Basic Auth credentials. This will allow to use a different set for credentials for the Basic Auth and Principal.
- Friendly error message with Exception if the Basic Auth fails.
- Additional test cases.

radcortez:
@jgallimore @exabrial I do have a question:

When using the Basic Auth without the Security Principal, what should be the value of SessionContext.getCallerPrincipal()? At the moment, it returns the Basic Auth login value. It seems the EJB context gets a Principal from this authentication, when want we were trying to achieve was only the Basic Auth. I'm not sure if this Principal should propagate to the EJB context.

radcortez:
Proposed documentation:
https://gist.github.com/radcortez/fb040e83761638ab40d8a19d06b8bfc3

radcortez:
I've pushed the documentation into my own git repo. Here is a PR:
https://github.com/radcortez/tomee-site/pull/1

Probably easier to review this way. Also easy to generate a patch.

PR #53, opened 2016-11-27
Support disabling of builtin TomEEJohnzonProvider and TomEEJsonpProvi…

robertpanzer:
…der.

At the moment the TomeeJohnzonProvider and TomeeJsonpProvider are always added are only taken out of the way for an application if it contains the JacksonJsonProvider.
However all other builtin JAX-RS providers can be disabled globally by the properties `<fully qualified class name>.activated = false`.

If you create an own JAX-RS provider (in this case a MessageBodyReader/Writer) that for example wraps the JacksonJsonProvider because it's not easily possible to configure it as a resource via an ObjectRecipe.
Then you have to start fiddling with provider priorities which are computed from the set of media types produced/consumed by a MessageBodyWriter/Reader and can no longer use the wildcard media type used originally by the JacksonJsonProvider.

Therefore this PR simply evaluates these properties for the 2 Johnzon providers as well.


robertpanzer:
This change was committed as 0b8ba18148ac86ec22ed352def3e570fe887b377

PR #55, opened 2017-01-04
backport cipher change from tomee 7

ivanjunckes:


jgallimore:
This is for https://issues.apache.org/jira/browse/TOMEE-1995. Many thanks for the fix!

PR #56, opened 2017-01-07
moviefun-rest example was not working with Tomee 7

ivanjunckes:
Fixed 2 things:
1 - Tables were not being generated automatically 
2 - JSON parsing in JAX-RS was not working as Tomee 1.7.X. The previous version returned a different payload comparing to Tomee 7. I did small adjustments in the front-end so it could work

I should follow up this PR with another change to add documentation for Intellij (dependent on moviefun-rest app) which is not up to date in the previous doc. 
http://tomee.apache.org/tomee-and-intellij.html


radcortez:
@ivanjunckes can you please review this PR?

PR #60, opened 2017-03-02
amq and ibm mq portability using tomee

suyogbarve:
amq and ibm mq portability using tomee


radcortez:
@jgallimore I know you were working on this some time ago. can you review it?

jgallimore:
Yes, I'll get that done today. Thanks for the nudge.

jgallimore:
@suyogbarve Thank you for putting this together, it looks great. I have a couple of points for feedback:

* Can you add the ASL header to your source files?
* Can ChatBean and ChatBeanTest be removed?
* Update Java version to 1.8 in pom.xml for maven-compiler-plugin
* Remove the JMS 2.0 API, and update org.apache.tomee:javaee-api to 8.0 (this covers JMS 2.0)
* Remove the janino dependency (unless its needed for something that I have missed)

You could also check out this JIRA: https://issues.apache.org/jira/browse/TOMEE-2162, and this commit: https://github.com/apache/tomee/commit/d9f3ca8498e1696637eac569f0396b63030703b0 - I enabled activations to be specified on the container in the application's resources.xml:

```
<Container id="mdb-container">
   activation.some.property = value
</Container>
```

that activation property will be applied to all the MDBs in that container - so quite handy if you need to specify the WMQ host, for example, on each MDB.

If memory serves me correctly, you can override these with system properties as well:

-D<mdb-container-id>.activation.<property> = value.

I'm happy to help with this, and we don't need all of these done to merge it in, but we would need the ASL header on the files first.

PR #64, opened 2017-03-16
Inconsistent timer view when using quartz persistence

bodop:
In TomEE timers are stored twice. Once in the TimerStore and once in the quartz scheduler. Quartz itself may use in memory or database storage. It is hard to keep both stores in sync, especially if database persistence is used in quartz and TomEE is restarted. This leads to inconsistencies.

This pull request tries to avoid the inconsistencies by removing one storage location (TimerStore).


rmannibucau:
I like the idea but think it doesnt work cause we need to handle transactionality there so we need to enhance TimerStore to rely more on quartz outside of transactions, no?

bodop:
I am not sure if I really understood the question. But I see three cases:

 1. Manipulation of persisted timers within a transaction. This should
    be handled perfectly using persistent quartz timers, since the
    database connection takes part in the transaction.
 2. Manipulation of persisted timers outside a transaction. This might
    fail. But I am not sure what the EJB specification says about this
    and if it even makes sense to do so.
 3. Manipulation of non persisted timers. Here my current patch seems to
    be overkill (unless if running in a cluster environment), since
    timers are persisted and deleted during restart. I like the idea of
    an independent Quartz-Scheduler for non-persistent timers.

rmannibucau:
You are kind of right except assuming we use a db which will not be the default (memory) so 1 is an issue in this case. Said otherwise i like the move of this pr but it would break ejb spec in most cases (few users rely on jdbc).

bodop:
OK, I do not remember excatly what I was doing in this pull request. But
it seems that the whole work is still done in
TimerService.transactionComplete(). So the main functionality is left
untouched. And I wonder why it works with
"org.apache.openejb.quartz.impl.jdbcjobstore.JobStoreTX" since the
transaction is already completed.

I think all test cases in container/openejb-core still work. It would be
interesting to find a test case that works in original openejb and not
in the pull request.

rmannibucau:
A user reported jobstoretx was not working and digging into it i realized it cant work until you hack the container so sounds unsane :s.

We also have some close tests due to oracle licensing so maybe coverage is not great on os side

PR #65, opened 2017-04-06
Refactored mockito integration

kalgon:
Refactored mockito integration which allows qualified mocks.
In the end, I had to keep MockitoInjector which is still needed for `@EJB` injections.


AndyGee:
@kalgon can you resolve the conflicts? I'll take a look tomorrow.

PR #66, opened 2017-05-02
[WIP] Fixing errors in Test

tbookless:
An equal sign was used in a few places instead of a colon.
Redundant escape file
32 unused imports


tbookless:
There are just too many issues with the tests to take a general approach. I don't think that it will get merged because the number of changes will be too, high. Going to start again with a more targeted approach.

PR #72, opened 2017-06-16
[TOMEE-2060] Make app naming context read only (switched on/off with property)

katya-stoycheva:
Initial proposal (there are things to be clarified)

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/assembler/classic/Assembler.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/assembler/classic/Assembler.java
@@ -269,6 +269,8 @@ public class Assembler extends AssemblerTool implements org.apache.openejb.spi.A
     public static final String TIMER_STORE_CLASS = "timerStore.class";
     private static final ReentrantLock lock = new ReentrantLock(true);
     public static final String OPENEJB_TIMERS_ON = "openejb.timers.on";
+    static final String FORCE_READ_ONLY_APP_NAMING = "openejb.forceReadOnlyAppNamingContext";
+
     public static final Class<?>[] VALIDATOR_FACTORY_INTERFACES = new Class<?>[]{ValidatorFactory.class};
     public static final Class<?>[] VALIDATOR_INTERFACES = new Class<?>[]{Validator.class};
     private final boolean skipLoaderIfPossible;
@@ -1030,15 +1032,12 @@ public class Assembler extends AssemblerTool implements org.apache.openejb.spi.A
 
                 systemInstance.fireEvent(new AssemblerAfterApplicationCreated(appInfo, appContext, allDeployments));
                 logger.info("createApplication.success", appInfo.path);
-                
-                //TODO: set proper default
-                //required by spec: EE.5.3.4, used together with tomcat#jndiExceptionOnFailedWrite 
-                if("true".equals(SystemInstance.get().getProperty("forceReadOnlyAppNamingContext", "true"))) {
-                    ensureAppNamingContextIsReadOnly(allDeployments);
-                    //TODO message
-                    logger.info("createApplication.success", appInfo.path);
-                }
 
+                //required by spec EE.5.3.4
+                if(setAppNamingContextReadOnly(allDeployments)) {
+        	        logger.info("createApplication.naming", appInfo.path);
+                }
+              
                 return appContext;
             } catch (final ValidationException | DeploymentException ve) {
                 throw ve;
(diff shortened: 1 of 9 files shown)


container/openejb-core/src/main/java/org/apache/openejb/assembler/classic/Assembler.java
```diff
@@ -1029,6 +1030,14 @@ private AppContext createApplication(final AppInfo appInfo, ClassLoader classLoa
 
                 systemInstance.fireEvent(new AssemblerAfterApplicationCreated(appInfo, appContext, allDeployments));
                 logger.info("createApplication.success", appInfo.path);
+                
+                //TODO: set proper default
+                //required by spec: EE.5.3.4, used together with tomcat#jndiExceptionOnFailedWrite 
+                if("true".equals(SystemInstance.get().getProperty("forceReadOnlyAppNamingContext", "true"))) {
```
rmannibucau: please prefix it openejb. (openejb.forceReadOnlyAppNamingContext)

also shouldnt the default be false to ensure it is writable by default (backward compatibility)

also think createApplication.success is not the right key yet ;)

container/openejb-core/src/main/java/org/apache/openejb/assembler/classic/Assembler.java
```diff
@@ -1049,6 +1058,19 @@ private AppContext createApplication(final AppInfo appInfo, ClassLoader classLoa
...
 
+    private void ensureAppNamingContextIsReadOnly(final List<BeanContext> allDeployments) {
+        for(BeanContext beanContext : allDeployments) {
+            Context ctx = beanContext.getJndiContext();
+            if(ctx instanceof IvmContext) {
+               ((IvmContext) ctx).setReadOnly(true);
+            } else if(ctx instanceof ContextHandler) {
+            ((ContextHandler)ctx).setReadOnly();
+            } else {
+                //TODO: log only?
```
rmannibucau: shouldn't be possible so can be ignored for now

container/openejb-core/src/main/java/org/apache/openejb/assembler/classic/Assembler.java
```diff
@@ -1049,6 +1058,19 @@ private AppContext createApplication(final AppInfo appInfo, ClassLoader classLoa
         }
     }
 
+    private void ensureAppNamingContextIsReadOnly(final List<BeanContext> allDeployments) {
+        for(BeanContext beanContext : allDeployments) {
+            Context ctx = beanContext.getJndiContext();
+            if(ctx instanceof IvmContext) {
+               ((IvmContext) ctx).setReadOnly(true);
+            } else if(ctx instanceof ContextHandler) {
+            ((ContextHandler)ctx).setReadOnly();
```
rmannibucau: side note on casting: on the list we got a discussion saying we prefer the ${type}.class.cast(instance) style instead of (($type) instance). Not a blocker for me but just mentionning it if you see that style in the code elsewhere.

container/openejb-core/src/main/java/org/apache/openejb/core/ivm/naming/IvmContext.java
```diff
@@ -399,7 +408,10 @@ public void destroySubcontext(final Name name) throws NamingException {
     }
 
     public Context createSubcontext(String name) throws NamingException {
-        checkReadOnly();
+        if(checkReadOnly()) {
+            //TODO: throw exception or return null? tomcat returns null
+            return null;
```
rmannibucau: null is fine if there is a one time - 10 calls will log a single time - log line (warning?)

container/openejb-core/src/main/java/org/apache/openejb/core/ivm/naming/IvmContext.java
```diff
@@ -477,9 +489,36 @@ public String getNameInNamespace() throws NamingException {
...
+     * jndiExceptionOnFailedWrite property is defined by tomcat and is used in similar context for web app components
+     * https://tomcat.apache.org/tomcat-7.0-doc/config/context.html#jndiExceptionOnFailedWrite
+     * 
+     */
+    protected boolean checkReadOnly() throws OperationNotSupportedException {
+      //TODO: should it log?
         if (readOnly) {
-            throw new OperationNotSupportedException();
+            //alignment with tomcat behavior
+            if("true".equals(System.getProperty("jndiExceptionOnFailedWrite"))) {
```
rmannibucau: SystemInstance.get().getProperty and openejb prefix as well, default should fail as before

container/openejb-core/src/main/java/org/apache/openejb/core/ivm/naming/IvmContext.java
```diff
@@ -477,9 +489,36 @@ public String getNameInNamespace() throws NamingException {
...
+                throw new OperationNotSupportedException();
+            }
+            return true;
+        }
+        return false;
+    }
+    
+    public void setReadOnly(boolean isReadOnly) {
+        //TODO: should it log?
```
rmannibucau: no need to log it until there is an issue with this being called (which is more in write methods)

rmannibucau:
applied, build in progress on https://ci.apache.org/builders/tomee-trunk-ubuntu/builds/778

katya-stoycheva:
Thank you for reviewing this. Three more questions/comments from my side:
- Is there Tomee code convention documented somewhere? I tried to stick to the original format but obviously the IDE has applied some own preferences and I saw check-formatting step is failing.
- About the arquillian tests - I have added a test to check only the default behavior since this is a config change and is applied on container start. (I didn't manage to change it via the test framework) How do you test different behavior that depends on a system property for example?  I was thinking of adding new container for each config set but it would require specifying different mvn profile on build and someone has to invoke these which could be seen as overhead. Do I need to add more arquillian tests for read-only context? 
- About logging (on your request: if there is a one time - 10 calls will log a single time - log line (warning?)) It would need to propagate and synchronize a counter among nested contexts and it would be unnecessary overhead in my opinion. But if you think this would be hepful, I can add it.

rmannibucau:
1. not that much, we are not that strong about it cause we all have different habits and thinking about it
2. you can create a new surefire execution for this test with another arquillian container config, that said the embedded testing in openejb-core (which forks by test) should be enough probably
3. you logged it was read only in the assembler, this is enough for now probably

AndyGee:
@katya-stoycheva can you please close this - it was applied. Thank you very much for your contribution :-)

PR #75, opened 2017-06-21
[TOMEE-2070] IvmContext close now checks for read only

katya-stoycheva:


rmannibucau:
Applied, thanks a lot for this follow up Katya!

PR #76, opened 2017-06-22
TOMEE-2073: The value of EJB.beanName() in some cases is ignored

SvetlinZarev:


SvetlinZarev:
It was merged, so closing the PR

PR #79, opened 2017-06-28
TOMEE-2081 Move the Mdb control logic from ActiveMQ to the MdbContainer

jgallimore:


container/openejb-core/src/main/java/org/apache/openejb/core/mdb/MdbContainer.java
```diff
@@ -505,4 +574,139 @@ public static BeanContext current() {
...
+        public ActivationSpec getActivationSpec() {
+            return activationSpec;
+        }
+
+        public boolean isStarted() {
+            return started.get();
+        }
+
+        public void start() throws ResourceException {
+            if (started.get()) {
```
rmannibucau: compareAndSet no? otherwise 2 threads can see it false and call activation which is likely not what we want. wdyt?

(indeed stop is consistent so has the same issue)

container/openejb-core/src/main/java/org/apache/openejb/core/mdb/MdbContainer.java
```diff
@@ -505,4 +574,139 @@ public static BeanContext current() {
...
+                return;
+            }
+
+            final ClassLoader oldCl = Thread.currentThread().getContextClassLoader();
+            Thread.currentThread().setContextClassLoader(classLoader);
```
rmannibucau: try/finally for such operation is important to not mess up a reused thread (spent 3 days to fix such an issue in our arquillian adapter so know it looks trivial to fix but hard to find when it pops up)

PR #80, opened 2017-06-28
TOMEE-2082 - TOMEE-1908 allow JCA connectors to cast to their connection impl

jgallimore:
Proposed backport of TOMEE-2082.


container/openejb-core/src/main/java/org/apache/openejb/resource/AutoConnectionTracker.java
```diff
@@ -25,6 +25,8 @@
 import org.apache.openejb.dyni.DynamicSubclass;
 import org.apache.openejb.util.LogCategory;
 import org.apache.openejb.util.Logger;
+import org.apache.openejb.util.proxy.LocalBeanProxyFactory;
+import sun.misc.Unsafe;
```
rmannibucau: guess you want to drop this one ;)

PR #81, opened 2017-06-29
[TOMEE-2083] List IvmContext should list all of its federated context bindings

katya-stoycheva:


jeanouii:
Thanks for the PR.
All applied to master

PR #82, opened 2017-07-03
Add missing test scope to cglib

SvetlinZarev:


jeanouii:
as discussed in the mailing list @rmannibucau and myself are ok with this PR.

PR #85, opened 2017-07-05
TOMEE-1974- basic authentication

jgallimore:
This was originally merged here https://git1-us-west.apache.org/repos/asf?p=tomee.git;a=tree;hb=e0397f49, and subsequently lost in the revert of 05 Dec 2016 (http://tomee-openejb.979440.n4.nabble.com/Commit-deletion-td4680672.html). Is there some potential to restore this lost set of commits?


rmannibucau:
@jgallimore don't recall where was the thread about it but think 2 inputs are important:

1. we already have the ability to use basic
2. user can always configure the desired security (even another one) creating physically tomee webapp and configuring the ejbdservlet with standard servlet security mecanisms
2.bis. you can also do it on the host directly if the instance = 1 webapp = ejbd

2 can need some more doc inputs indeed but I'd prefer to not introduce a new mecanism for 1 and to not do something specific for that need.

PR #91, opened 2017-07-13
Resource pre destroy master

jgallimore:


container/openejb-core/src/main/java/org/apache/openejb/assembler/classic/Assembler.java
```diff
@@ -2469,36 +2469,40 @@ public void destroyApplication(final AppInfo appInfo) throws UndeployException {
     }
 
     private void destroyLookedUpResource(final Context globalContext, final String id, final String name) throws NamingException {
+        // check to see if the resource is a LazyObjectReference that has not been initialized
+        // if it is, we'll remove the LazyObjectReference, rather than do a lookup causing it
+        // to be instantiated
+        final String ctx = name.substring(0, name.lastIndexOf("/"));
```
AndyGee: I'd like to see more safety here, that lastIndexOf isn't -1 etc. I know it wasn't there before, but just seems to be worth now you're on it.

container/openejb-core/src/test/java/org/apache/openejb/assembler/classic/ApplicationResourceLifecycleTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+public class ApplicationResourceLifecycleTest {
+
+    private static final AtomicBoolean POST_CONSTRUCT = new AtomicBoolean(false);
+    private static final AtomicBoolean PRE_DESTROY = new AtomicBoolean(false);
+
```
AndyGee: Would like to see Atomics set in an @BeforeClass

jeanouii: Can you elaborate Andy?
I see the atomics to track if the PreDestroy or PostConstruct methods are going to be called in the ApplicationResource inner class.

There is only one test method.

PR #104, opened 2017-10-02
HTTP(s) basic auth failed if password contained ampersand passed via …

pendor:
…basic.password URL parameter

A double-decode bug caused URLDecode to be applied twice to parameters passed in
via URL including basic.username and basic.password.  The parameters were automatically
decoded by the call to URI.getQuery() then again as each parameter was parsed and added
to the returned Map in MulticastConnectionFactory.URIs.parseQuery().  parseQuery() splits the
query string on the ampersand character then explictly URLDecode's each value.  Since
URI.getQuery() had already decoded the basic.password parameter, the splitting process
in parseQuery truncated the password at the first ampersand character.

Instead, URI.getRawQuery() should be called to get the still URLEncoded query string.  The
splitting and subsequent decoding in parseQuery() then correctly extracts the full password
from the query string.

PR contains failing unit test & fix.

### DIFF
--- a/server/openejb-client/src/main/java/org/apache/openejb/client/MulticastConnectionFactory.java
+++ b/server/openejb-client/src/main/java/org/apache/openejb/client/MulticastConnectionFactory.java
@@ -133,7 +133,7 @@ public class MulticastConnectionFactory implements ConnectionFactory {
         }
 
         public static Map<String, String> parseParamters(final URI uri) throws URISyntaxException {
-            return uri.getQuery() == null ? new HashMap<String, String>(0) : parseQuery(stripPrefix(uri.getQuery(), "?"));
+            return uri.getRawQuery() == null ? new HashMap<String, String>(0) : parseQuery(stripPrefix(uri.getRawQuery(), "?"));
         }
 
         public static String stripPrefix(final String value, final String prefix) {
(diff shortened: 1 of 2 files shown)


rzo1:
I can follow your explanation. I took a quick look into `MulticastConnectionFactory` and confirm, that the decoding is happening twice. I think, that we do not have a related JIRA ticket for this issue.

I think, that we can enhance the test-case with `try-with-resources` and without the `e.printStacktrace()` statements inside the test-case.

WDYT @jeanouii ?

rzo1:
Shall we create a JIRA-Ticket for this? @jeanouii

rzo1:
I created https://issues.apache.org/jira/browse/TOMEE-2656 for this.

Did not get an answer for the comment:

> I think, that we can enhance the test-case with try-with-resources and without the e.printStacktrace() statements inside the test-case.

WDYT @cesarhernandezgt  ?

jgallimore:
I think this looks good to merge. I'll give it a quick test and get it merged.

jgallimore:
Merged and backported to 7.1.x and 7.0.x. Thanks for the PR!

rzo1:
I closed the related JIRA and set the fix version(s)

jgallimore:
@rzo1 Many thanks for the review, the JIRA admin and for pushing this through, its really appreciated.

PR #110, opened 2017-11-07
Fixed Ejb Proxy Handler not to unmarshall for every remote call

tcanavan:
TOMEE-2137: Reintroduced optimisation

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/core/ivm/BaseEjbProxyHandler.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/core/ivm/BaseEjbProxyHandler.java
@@ -313,7 +313,7 @@ public abstract class BaseEjbProxyHandler implements InvocationHandler, Serializ
                 final SecurityService securityService = SystemInstance.get().getComponent(SecurityService.class);
                 securityService.associate(localClientIdentity);
             }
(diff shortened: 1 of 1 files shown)


rmannibucau:
-1 this breaks the case the api are deployed in different version/classloader. The check should rather be enriched but probably not deleted

tcanavan:
Not familiar with the source code on how to achieve that. Any ideas would be great

tcanavan:
public static boolean isIntraVmCopyOperation() {
        return state() == State.COPY;
    }

    public static boolean isStatefulPassivationOperation() {
        return state() == State.PASSIVATION;
    }

    public static boolean isCrossClassLoaderOperation() {
        return state() == State.CLASSLOADER_COPY;
    }

CLASSLOADER_COPY seems to cover that situation. So why the extra condition for remote interfaces

rmannibucau:
Because it is needs to be automatic and it is a global setting so you need to be able to not use it by default but enforce it for the API which require it.

tcanavan:
Ok I understand you wan to apply the optimisation on a pro interface basis. Would this happen in some xml description like it does in other containers?

rmannibucau:
There are a few points:

1. we can make the option configurable in openejb-jar.xml properties if we want (optional)
2. we must enforce the the case if it will fail cause of the classloader "conflict" (which is why this test was added), silently, without any config
3. we should probably log a warning in this case (once, then we cache in BeanContext.data it was already logged to avoid to pollute logs too much)

wdyt?

PR #118, opened 2018-01-12
[BACKPORT] Improves MDB performance with instance pool

danielsoro:


tveronezi:
Hi @danielsoro 
We are going to use this PR instead https://github.com/apache/tomee/pull/119
This way we will be able to track the original changes. Can you close this one?

PR #122, opened 2018-03-05
TomEE MicroProfile dist with Config

radcortez:


rmannibucau:
applied with an upgrade to g-config to ensure we pass tcks

radcortez:
Thanks

PR #134, opened 2018-07-13
JndiEncBuilder missing ThreadLocalContextManager.APPLICATION case + swagger example

tveronezi:


tveronezi:
This is a work in progress. I will create another PR once everything is done.
[]s,
Thiago.

PR #135, opened 2018-07-16
[TOMEE-2201] NPE when injecting Application

tveronezi:
There are two things here:
* the `javax.ws.rs.core.Application` option was not in the if chain in JndiEncBuilder.java.
* MapObjectReference.java was throwing NPE if the referenced map was null. 

Help on creating unit tests for these guys is needed.


tveronezi:
CustomContext testing is already covered by server/openejb-cxf-rs/src/test/java/org/apache/openejb/server/cxf/rs/CustomContextTest.java

PR #141, opened 2018-07-31
TOMEE-2209 - TomEE 8 Microprofile 1.2 Support

radcortez:
This for now includes specs up to MP 1.2. I'll keep working on this to add specs to push for MP 1.3.


jeanouii:
@radcortez Thanks Roberto.
What is the status for this?
Is it ready to get reviewed?

radcortez:
I believe so. This has all the integration for MP 1.2.

A couple of tests fail due to https://github.com/apache/tomee/pull/142. We need to decide how to move forward with that.

PR #166, opened 2018-09-15
Apply Jonathan Gallimore's JMX Patch to 7.0.5

exabrial:
Backport this to 7.0.x

https://github.com/apache/tomee/pull/160/files

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/assembler/classic/Assembler.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/assembler/classic/Assembler.java
@@ -109,6 +109,7 @@ import org.apache.openejb.loader.ProvisioningUtil;
 import org.apache.openejb.loader.SystemInstance;
 import org.apache.openejb.monitoring.DynamicMBeanWrapper;
 import org.apache.openejb.monitoring.LocalMBeanServer;
+import org.apache.openejb.monitoring.MBeanPojoWrapper;
 import org.apache.openejb.monitoring.ObjectNameBuilder;
 import org.apache.openejb.monitoring.remote.RemoteResourceMonitor;
 import org.apache.openejb.observer.Observes;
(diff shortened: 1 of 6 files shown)


exabrial:
Nuts, this PR breaks hawt.io FYI

jgallimore:
Interesting - you got any more details? I'm happy to take a look.

exabrial:
Yes, to reproduce:

1. Build this branch from current master
1. Unzip the plume tar.gz
1. Download https://oss.sonatype.org/content/repositories/public/io/hawt/hawtio-default/2.1.0/hawtio-default-2.1.0.war to webapps/hawtio.war [the filename hawtio.war is important, must be exactly that]
1. start the server with catalina.sh run
1. navigate to localhost:8080/hawtio and see jmx is working
1. stop the server
1. deploy the tomee.xml I emailed to you
1. restart the server
1. nagivate to localhost:8080/hawtio and see the jmx section is missing

exabrial:
It seems to revolve around the setting of a remote broker rather than using a local broker. The tomee.xml I sent you has a remote broker and breaks hawtio, but if you use a vm://localhost it works fine. Strange

exabrial:
Hey I think I know what's causing the problem...it looks like the patch set we applied causes a `NullPointerException` when clicking on the AMQ RA. 

Hawtio or jolokia might not be able to handle that.

screen shot 2018-09-19 at 4 07 23 pm

exabrial:
It looks like on the ActiveMQ resource adapter, `mBeanInfo.getAttributes()` is returning null, breaking hawtio. They should obviously handle this case, but also guessing we're probably not supposed to be doing that either: 
screen shot 2018-09-23 at 9 53 58 pm

jgallimore:
@exabrial - fixed here: https://github.com/apache/tomee/commit/e1853b10c52f72c7e9dd43c049098d7f69a43a49 thanks for the report!

exabrial:
Thanks for the fix! I'm surprised the jvm allowed the nulls.

PR #170, opened 2018-09-22
[7.0.x Branch] TOMEE-2240 ManagedScheduledExecutorService doesn't allow setting max value

exabrial:
Fix https://issues.apache.org/jira/projects/TOMEE/issues/TOMEE-2240


exabrial:
> While this class inherits from ThreadPoolExecutor, a few of the inherited tuning methods are not useful for it. In particular, because it acts as a fixed-sized pool using corePoolSize threads and an unbounded queue, adjustments to maximumPoolSize have no useful effect

need to read the manual better

PR #171, opened 2018-09-22
[Master Branch] TOMEE-2240 ManagedScheduledExecutorService doesn't allow setting max value

exabrial:
Fix https://issues.apache.org/jira/projects/TOMEE/issues/TOMEE-2240


rmannibucau:
Isnt 5s too small?
Also, dont forget to add new params to service-jar.xml in openejb-core.
Otherwise looks good!

exabrial:
Actually I need to close this! I wrote this patch before I read the javadoc on the scheduled executor service thoroughly. The max setting is a little misleading. When you wrote the original code you left the max out because it doesn't have the intended effect

rmannibucau:
No issue, thanks for taking time to PR and recheck it.

PR #175, opened 2018-09-29
[tomee-7.0.x][TOMEE-2249] Fix EclipseLink NPE on init

exabrial:


exabrial:
Fix TOMEE-2249 https://issues.apache.org/jira/projects/TOMEE/issues/TOMEE-2249?filter=reportedbyme by implementing required interfaces and executing method

container/openejb-jpa-integration/src/main/java/org/apache/openejb/jpa/integration/eclipselink/MBeanOpenEJBRuntimeServices.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.openejb.jpa.integration.eclipselink;
+
+import org.eclipse.persistence.internal.sessions.AbstractSession;
+import org.eclipse.persistence.sessions.Session;
+
+public class MBeanOpenEJBRuntimeServices extends OpenEJBRuntimeServices implements MBeanOpenEJBRuntimeServicesMBean {
```
rmannibucau: ideally we would have a single impl, cant we reuse one existing impl since they do nothing here?

only PLATFORM_NAME is set and it is not really needed since only an error thing

container/openejb-jpa-integration/src/main/java/org/apache/openejb/jpa/integration/eclipselink/OpenEJBServerPlatform.java
```diff
@@ -63,4 +65,11 @@ protected void registerSynchronization_impl(AbstractSynchronizationListener list
             transaction.registerInterposedSynchronization(synchronization);
         }
     }
+
+        @Override
+        public void prepareServerSpecificServicesMBean() {
+            if (getDatabaseSession() != null && shouldRegisterRuntimeBean) {
+                 this.setRuntimeServicesMBean(new MBeanOpenEJBRuntimeServices(getDatabaseSession()));
```
rmannibucau: ensure it uses mBeanServer which has a toggle "skip jmx" on the instance

container/openejb-jpa-integration/src/main/java/org/apache/openejb/jpa/integration/eclipselink/MBeanOpenEJBRuntimeServices.java
exabrial: I certainly can use the Glassfish one here if you would like. But, to remain consistent with how all the other ones are implemented, I would suggest we don't.

screen shot 2018-10-08 at 1 18 15 pm

container/openejb-jpa-integration/src/main/java/org/apache/openejb/jpa/integration/eclipselink/OpenEJBServerPlatform.java
exabrial: Good idea, I added a check there

exabrial:
@rmannibucau If this looks good, could you merge?

container/openejb-jpa-integration/src/main/java/org/apache/openejb/jpa/integration/eclipselink/MBeanOpenEJBRuntimeServices.java
rmannibucau: Was more thinking about dropping this and keeping only OpenEJBRuntimeServices

container/openejb-jpa-integration/src/main/java/org/apache/openejb/jpa/integration/eclipselink/OpenEJBServerPlatform.java
```diff
@@ -26,13 +27,14 @@
...
+public class OpenEJBServerPlatform extends JMXServerPlatformBase implements JMXEnabledPlatform {
     public OpenEJBServerPlatform(final DatabaseSession newDatabaseSession) {
         super(newDatabaseSession);
         try {
             mBeanServer = MBeanServer.class.cast(
                 OpenEJBServerPlatform.class.getClassLoader().loadClass("org.apache.openejb.monitoring.LocalMBeanServer")
                     .getMethod("get").invoke(null));
+            this.prepareServerSpecificServicesMBean();
         } catch (final Exception e) {
             // no-op
```
rmannibucau: Fail instead of noop?

exabrial: I agree with you; I don't understand why they split it out either, but they literally did it for _every other server_ so there's probably a good reason. 

On the fail vs noop, that was there before me. I suggest leaving it so we don't change current behavior (if JMX fails to initialize, we still alow the application to run. Right now JMX is failing.)

> 6e2a4f7c419 (Thiago Veronezi    2015-11-23 14:38:43 -0500 36)                     .getMethod("get").invoke(null));
> 515306f7ba0 (Jonathan S. Fisher 2018-09-29 13:10:49 -0500 37)             this.prepareServerSpecificServicesMBean();
> 6e2a4f7c419 (Thiago Veronezi    2015-11-23 14:38:43 -0500 38)         } catch (final Exception e) {
> 6e2a4f7c419 (Thiago Veronezi    2015-11-23 14:38:43 -0500 39)             // no-op
> 6e2a4f7c419 (Thiago Veronezi    2015-11-23 14:38:43 -0500 40)         }
> 6e2a4f7c419 (Thiago Veronezi    2015-11-23 14:38:43 -0500 41)     }
> 6e2a4f7c419 (Thiago Veronezi    2015-11-23 14:38:43 -0500 42)

PR #176, opened 2018-10-09
TOMEE-2253 - tomee.sh -version not working properly with Java 11

danielsoro:


danielsoro:
@rmannibucau  What do you want mean: should likely be closed?

rmannibucau:
@danielsoro classloader.close() must be called to not leak and enable to read files after the end of the container.

danielsoro:
@rmannibucau hmm.. my question is: If we need to call close, why are not we doing it when using CustomizableURLClassLoader? is it means which we need to do the same for CustomizableURLClassLoader?

rmannibucau:
Java 6 didnt have that method but we should close all our classloaders once yes.

danielsoro:
Hey @rmannibucau,

I updated it to call `close` and also now we call the ClassLoader.registerAsParallelCapable(); in a static block.

hashcode and equals still using the super, the new class doesn't have any property which we can use to create a custom equals ou hashcode. If you have some idea how can I create ones, please show off.

rmannibucau:
well some thoughts:

1. most of the time we dont need this dynamism so maybe we can just rework the distro and delete it
2. we already hack equals/hashcode in cxf container loader but it is not super robust (this is why we have a flag to disable it) so maybe not the best solution but current impl clearly breaks some apps
3. you can open java modules to get the URLClassPath, even of the system classloader in j11 so maybe we just document how to open it and replace the cast by some reflection?

danielsoro:
@rmannibucau, 

TomEE is not showing any exception when running with j11 (I was able to start and deploy a simple application on it really fine), that is not the issue.

It just appears with tomee.sh when I'm using the CLI commands. 
Maybe applying that change just to cover the org.apache.openejb.cli.Bootstrap should works.

WDYT?

rmannibucau:
Yes, this is what i had in mind as well. That said if it is just for our CLI your hack is fine, we should just lock it for other cases like openejb-standalone

danielsoro:
@rmannibucau,

Nice.. working on it. 
Thank you! :)

danielsoro:
Hi @rmannibucau,

patch updated. It seems to look better now.
Please, review it and thank you for all the help, I really appreciate it.

rmannibucau:
Hope i didnt misread the diff (starts to be late ;)) but it misses a close and the urlclassloader cast will not work on all java versions right?

danielsoro:
Ohh.. sorry I sent the wrong patch. 
Pushing the correct one.

danielsoro:
Ok @rmannibucau,

sent the correct patch, let me know what do you think.
Thank you!

rmannibucau:
Hmm, you close the classloader before it is used so it is no more usable normally - or it will leak.

Side question: any issue having an impl os ClassPath using the classloader to keep existing one for java 8 users?

danielsoro:
@rmannibucau,

Yeah, I'm closing it before, going to fix it. 
I didn't get your question about j8's user, but it should not have an impact on j8's user.

rmannibucau:
@danielsoro if you change the classloader it can impact the users relying on that (it can be the case if they tune java or use system classloader only libs. This is rare but already saw it. This is why i think it can be worth not changing the default behavior.

danielsoro:
@rmannibucau I get it, but we isolate it just for the CLI. 

What is the scenario which you are seeing which can break something with that change? It should not have any impact on the application deployed in TomEE.

rmannibucau:
Thnking to users having a java -cp command launching Cli (we have a few).

Side note: i think you still close the classloader before it is finished to be used - after cli execution pby

jgallimore:
I think adding the test anyway is a good call.

tomee/apache-tomee/src/test/java/org/apache/tomee/TomEECliIT.java
```diff
@@ -0,0 +1,66 @@
...
+        });
+
+        final File tomee = (null != files ? files[0] : null);
+        if (tomee == null) {
+            fail("Failed to find Tomcat directory required for this test - Ensure you have run at least the maven phase: mvn process-resources");
+        }
+
+
+        final ProcessBuilder builder = new ProcessBuilder()
+                .command("java", "-cp", jar.getAbsolutePath() + ":" +
```
jgallimore: Using : won't work on Windows.

tomee/apache-tomee/src/test/java/org/apache/tomee/TomEECliIT.java
```diff
@@ -0,0 +1,66 @@
...
+                                tomee.getAbsolutePath() + "/lib/openejb-core-8.0.0-SNAPSHOT.jar:" +
+                                tomee.getAbsolutePath() + "/lib/commons-cli-1.2.jar",
+                        "org.apache.openejb.cli.Bootstrap", "classloadertest");
+
+        final Process start = builder.start();
+        start.waitFor();
+
+        final String result = IO.slurp(start.getInputStream());
```
jgallimore: I suspect the logic that creates this output is in the JAR added in test/resources, making it a bit invisible. Can you assemble it with ShrinkWrap instead?

tomee/apache-tomee/src/test/java/org/apache/tomee/TomEECliIT.java
```diff
@@ -0,0 +1,66 @@
...
+        if (tomee == null) {
+            fail("Failed to find Tomcat directory required for this test - Ensure you have run at least the maven phase: mvn process-resources");
+        }
+
+
+        final ProcessBuilder builder = new ProcessBuilder()
+                .command("java", "-cp", jar.getAbsolutePath() + ":" +
+                                tomee.getAbsolutePath() + "/lib/openejb-core-8.0.0-SNAPSHOT.jar:" +
+                                tomee.getAbsolutePath() + "/lib/commons-cli-1.2.jar",
+                        "org.apache.openejb.cli.Bootstrap", "classloadertest");
```
jgallimore: Add another test with a directory as a classpath element.

jeanouii:
@danielsoro  and all
Where are we at?
Is it ready to go?

danielsoro:
Hey @jeanouii, I've push a update for it
Just need the review.

PR #201, opened 2018-11-19
TOMEE-2278 - Use Managed Executor with Safeguard Fault Tolerance lib

brunobat:
Well... when we include the SafeguardInterceptor, Our alternative is not picked up on the test.
We still need to figure out a CDI extension to activate the code on TomEE, but first, the test has to work on it's own.


brunobat:
This is still not ready. The new Lib code is loaded but the bean override is not happening yet.

radcortez:
+1

danielsoro:
+1

PR #214, opened 2018-11-21
 TOMEE-2283 - Websocket tls basic auth example

brunobat:
This is still a work in progress.


brunobat:
Hi @rmannibucau ,
Thanks for your comments. 
I've performed the following changes:
* I've removed the NOTICE/LICENSE files.
* Added missing headers to a few files. 
* I updated the package name... I had forgot to rename the package name... :)
* removed value= when 
* correct, the deplyment name is not needed here. Removed.

In relation to the server.xml and the keystore files...
The setup is on purpose. My initial idea was to have an arquillian configuration that can be easily migrated to an actual server. This way, The user can get inspiration to update their configuration files.
The keystore can be generated, but I fill it will complicate the example's interpretation. I feel it's easier like this.

PR #218, opened 2018-11-23
TOMEE-2290 - Metrics timer example

ivanjunckes:


pom.xml
```diff
@@ -208,7 +208,7 @@
     <microprofile.health.version>1.0</microprofile.health.version>
     <microprofile.health.impl.version>1.0.0</microprofile.health.impl.version>
     <microprofile.metrics.version>1.1</microprofile.metrics.version>
-    <microprofile.metrics.impl.version>1.0.0</microprofile.metrics.impl.version>
+    <microprofile.metrics.impl.version>1.0.1-SNAPSHOT</microprofile.metrics.impl.version>
```
brunobat: Hi @ivanjunckes I get an error here:
`[ERROR] Failed to execute goal on project tomee-microprofile-webapp: Could not resolve dependencies for project org.apache.tomee:tomee-microprofile-webapp:war:8.0.0-SNAPSHOT: Failure to find org.apache.geronimo:geronimo-metrics:jar:1.0.1-SNAPSHOT in https://repository.apache.org/service/local/repositories/orgapachegeronimo-1067/content/ was cached in the local repository, resolution will not be reattempted until the update interval of Geronimo OpenAPI Staging has elapsed or updates are forced`
The lib is not found in the maven repo.

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
```diff
@@ -0,0 +1,151 @@
...
+                .addAsWebInfResource(new StringAsset("<beans/>"), "beans.xml");
+        return webArchive;
+    }
+
+    @ArquillianResource
+    private URL base;
+
+    @Test
+    public void testTimedMetric() {
+        WebClient.create(base.toExternalForm())
```
rmannibucau: maybe use Client (jaxrs) instead of a cxf api - without forgetting to close the client ;)

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
```diff
@@ -0,0 +1,151 @@
...
+                "    \"reusable\": false,\n" +
+                "    \"tags\": \"\",\n" +
+                "    \"type\": \"timer\",\n" +
+                "    \"typeRaw\": \"TIMER\",\n" +
+                "    \"unit\": \"nanoseconds\"\n" +
+                "  }\n" +
+                "}";
+
+        JsonObject expectedJson = Json.createReader(new StringReader(expected)).readObject();
+        assertEquals(expectedJson, metadataJson);
```
rmannibucau: this is not stable so either assert using keys (keys1.size() == keys2.size() && for all key value ==) or using another way to write the assertEquals

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
```diff
@@ -0,0 +1,151 @@
...
+        assertTrue(metric.contains("# TYPE application:weather_day_status_seconds timer"));
+        assertTrue(metric.contains("application:weather_day_status_seconds{quantile=\"0.98\"}"));
+        assertTrue(metric.contains("# TYPE application:weather_day_status_seconds timer"));
+        assertTrue(metric.contains("application:weather_day_status_seconds{quantile=\"0.99\"}"));
+        assertTrue(metric.contains("# TYPE application:weather_day_status_seconds timer"));
+        assertTrue(metric.contains("application:weather_day_status_seconds{quantile=\"0.999\"}"));
+    }
+
+    private void assertJsonFormat(final String metricPath) {
+        final String metric = WebClient.create(base.toExternalForm())
```
rmannibucau: maybe load it as JsonObject and assert keys(), will make it inline and more reasable

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
```diff
@@ -0,0 +1,151 @@
...
+        assertJsonFormat(metricPath);
+    }
+
+    private void assertPrometheusFormat(final String metricPath) {
+        final String metric = WebClient.create(base.toExternalForm())
+                .path(metricPath)
+                .accept(MediaType.TEXT_PLAIN)
+                .get(String.class);
+
+        assertTrue(metric.contains("# TYPE application:weather_day_status_seconds summary timer"));
```
rmannibucau: as a general rule put a small description and the actual value in "message" parameter of the assert

proposal:

     Stream.of("expected text 1", "expected text 2")
        .forEach(text -> assertTrue("Expected: " + text + " to be present in " + metric, metric.contains(text)));

Also for an example you don't need to assert that much so just checking "application:weather_day_status_seconds_count 1.0" can be sufficient for the demonstration IMHO

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
ivanjunckes: What you mean by "Use the Client (jaxrs) instead of a cxf api?"?

rmannibucau: WebClient is a CXF API, you can use the standard JAXRS Client https://docs.oracle.com/javaee/7/api/javax/ws/rs/client/Client.html

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
ivanjunckes: I think keeping extra tests will not hurt :)

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
ivanjunckes: Done

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
ivanjunckes: Done

pom.xml
ivanjunckes: Fixed

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
ivanjunckes: But added the suggestion using Stream

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
```diff
@@ -157,7 +159,10 @@ private void assertJsonFormat(final String metricPath) {
         };
 
         Stream.of(expected)
-                .forEach(text -> assertTrue("Expected: " + text + " to be present in " + metric, metric.contains(text)));
+                .forEach(text ->
+                        assertTrue(
+                                "Expected: " + text + " to be present in " + metric,
+                                expectedJson.getJsonObject("weather_day_status").get("count") != null));
```
rmannibucau: s/count/text/ ;)

examples/mp-metrics-timed/pom.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rmannibucau: header ;)

ivanjunckes: Done :) Tks

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
ivanjunckes: Done :)

ivanjunckes:
@brunobat your  branch seems to be old dated. Can you update it and try again?

examples/mp-metrics-timed/src/test/java/WeatherServiceTest.java
ivanjunckes: Done

PR #219, opened 2018-11-26
Avoid String in a loop

otaviojava:
The reason to prefer StringBuilder is that both + and concat create a new object every time you call them (provided the right-hand side argument is not empty). This can quickly add up to a lot of objects, almost all of which are completely unnecessary.

```java
public class Main
{
    public static void main(String[] args)
    {
        long now = System.currentTimeMillis();
        slow();
        System.out.println("slow elapsed " + (System.currentTimeMillis() - now) + " ms");

        now = System.currentTimeMillis();
        fast();
        System.out.println("fast elapsed " + (System.currentTimeMillis() - now) + " ms");
    }

    private static void fast()
    {
        StringBuilder s = new StringBuilder();
        for(int i=0;i<100000;i++)
            s.append("*");      
    }

    private static void slow()
    {
        String s = "";
        for(int i=0;i<100000;i++)
            s+="*";
    }
}
```

* slow elapsed 11741 ms
* fast elapsed 7 ms

Also, this PR avoids unnecessary call in StringBuilder

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/config/AnnotationDeployer.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/config/AnnotationDeployer.java
@@ -2939,12 +2939,14 @@ public class AnnotationDeployer implements DynamicDeployer {
                         }
 
                         if (interfaces.size() != 1) {
-                            String msg = "When annotating a bean class as @MessageDriven without declaring messageListenerInterface, the bean must implement exactly one interface, no more and no less. beanClass=" + clazz.getName() + " interfaces=";
+                            StringBuilder msg = new StringBuilder("When annotating a bean class as @MessageDriven without" +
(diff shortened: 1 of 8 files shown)


container/openejb-core/src/main/java/org/apache/openejb/config/AnnotationDeployer.java
```diff
@@ -2939,12 +2939,14 @@ public EjbModule deploy(final EjbModule ejbModule) throws OpenEJBException {
                         }
 
                         if (interfaces.size() != 1) {
-                            String msg = "When annotating a bean class as @MessageDriven without declaring messageListenerInterface, the bean must implement exactly one interface, no more and no less. beanClass=" + clazz.getName() + " interfaces=";
+                            StringBuilder msg = new StringBuilder("When annotating a bean class as @MessageDriven without" +
```
danielsoro: What do you think to use String.format in cases like that?

PR #220, opened 2018-11-26
TOMEE-2292 - Config Microprofile Example

danielsoro:


examples/mp-config-example/pom.xml
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
```
brunobat: Hi Daniel, 
Thanks for your PR.
Can you please make the example work without a Parent definition? Please check: https://github.com/apache/tomee/blob/master/examples/README.adoc

PR #223, opened 2018-11-30
TOMEE-2301 - Concurrency utilities example

brunobat:
THIS IS A WORK IN PROGRESS


brunobat:
This is still a work in progress

examples/concurrency-utils/src/main/java/org/superbiz/executor/ManagedService.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Executes an opperation asynchronously, in a different thread provided by the {@link ManagedExecutorService}.
+     * The computation will carry on after the return of the method.
+     *
+     * @param value The demo data.
+     * @return A {@link CompletableFuture} that will return immediately.
+     */
+    public CompletableFuture<Integer> asyncTaskWithException(final int value) {
```
cesarhernandezgt: It would be nice to print also a message like in the asyncTask.
I also need to perform this on the example I'm running. I think it provides more visibility if the reader executes everything from command line.

examples/concurrency-utils/pom.xml
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.superbiz</groupId>
+    <artifactId>concurrency-utils</artifactId>
```
cesarhernandezgt: Does concurrency-utils actually involve future?
When I read concurrency I think in something like this: https://docs.oracle.com/javase/tutorial/essential/concurrency/index.html

My 2 cents fro the name:  asynch-utils

examples/concurrency-utils/src/main/java/org/superbiz/executor/ManagedService.java
brunobat: Will do, it's a good tip. Thanks!

ivanjunckes:
Hey @brunobat, tried  to build the project here but got a maven failure. I am using 3.5 and the project in enforcing 3.7.

You should follow the maven.version established in the project.
https://github.com/apache/tomee/blob/3f07dc964577b52a177173b6cfafbae5aa85adc0/pom.xml#L147

examples/concurrency-utils/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.arquillian.container</groupId>
+            <artifactId>arquillian-container-test-api</artifactId>
+            <version>1.4.0.Final</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
```
ivanjunckes: This is not needed. Add this to the pom file:
 <parent>
        <artifactId>examples</artifactId>
        <groupId>org.apache.tomee</groupId>
        <version>8.0.0-SNAPSHOT</version>
    </parent>

brunobat: Sorry @ivanjunckes what part is not needed?

ivanjunckes: maven-compiler-plugin, just add the xml for the parent.

jeanouii:
Hi all,

How are we looking on this one?
Is it ready to go?

brunobat:
Hi @jeanouii  
I'm still working on some additional examples. 
Will send an email to the list when ready.

PR #238, opened 2018-12-04
[IMPROVEMENT] TOMEE-2313 - Fixed surefire and openjpa versions on webapp archetype …

danielsoro:
…resource pom

Fixed OPENJB replace and add improvement to dependencies version for OPENJPA and JAVA EE API. :)


cesarhernandezgt:
nice work @danielsoro , do you think `maven-surefire-plugin` would also need this improvement?

danielsoro:
@cesarhernandezgt, well.. I believe which we don't need to have it sync with version used on TomEE.
Just the others make sense. It will depends of the maven version that user are using. The current version that we are using on archetype is good enough for surfire plugin.

PR #239, opened 2018-12-05
TOMEE-2321-exemple-mvc-cxf

Daniel-Dos:
I made a simple example with MVC 1.0 using Apache CXF as it was discussed in the email list that it would be interesting to have one.

https://issues.apache.org/jira/browse/TOMEE-2321


jgallimore:
I'm just looking through this now. Thanks for the contribution @Daniel-Dos !! Could you add the ASL headers to the new files?

Daniel-Dos:
I can add yes.
thank you.

examples/mvc-cxf/src/main/resources/META-INF/persistence.xml
```diff
@@ -0,0 +1,22 @@
...
+      <non-jta-data-source>mvc-demo</non-jta-data-source>
+
+      <class>org.superbiz.model.Address</class>
+      <class>org.superbiz.model.Person</class>
+
+      <properties>
+          <property name="eclipselink.target-database" value="org.eclipse.persistence.platform.database.H2Platform"/>
+          <property name="eclipselink.ddl-generation" value="create-tables"/>
+          <property name="eclipselink.ddl-generation.output-mode" value="database"/>
+      </properties>
```
jgallimore: Can you try adding this:

`<property name="openjpa.jdbc.SynchronizeMappings" value="buildSchema(ForeignKeys=true, Indexes=false, IgnoreErrors=true)"/>`

That should enable schema generation on OpenJPA (Webprofile and Plus) as well as EclipseLink (Plume).

Daniel-Dos: done .

PR #262, opened 2018-12-07
TOMEE-2337 - Code cleanup openejb container loader

cicekhayri:
Added JavaDoc's, use diamond operator, isEmpty method, Switch case statement and added @Override

### DIFF
--- a/container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
+++ b/container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
@@ -28,6 +28,11 @@ import java.util.Locale;
 
 public abstract class BasicURLClassPath implements ClassPath {
 
+    /**
+     * Returns the context ClassLoader for this Thread or null
+     * 
+     * @return ClassLoader ClassLoader
+     */
     public static ClassLoader getContextClassLoader() {
         return AccessController.doPrivileged(new PrivilegedAction<ClassLoader>() {
             @Override
@@ -41,7 +46,9 @@ public abstract class BasicURLClassPath implements ClassPath {
     private boolean ucpFieldErrorLogged;
 
     /**
-     *
+     * Add Jar to the URLClassPath
+     * 
+     * Throws Exception if fails any
      * @param jar URL
      * @param loader URLClassLoader
      * @throws Exception
@@ -75,7 +82,8 @@ public abstract class BasicURLClassPath implements ClassPath {
     }
 
     /**
-     *
+     * Adds Jars to the URLClassPath
+     * 
      * @param dir File
      * @param loader URLClassLoader
      * @throws Exception
@@ -125,7 +133,7 @@ public abstract class BasicURLClassPath implements ClassPath {
     }
 
     /**
-     *
+     * 
      * @param loader URLClassLoader
      * @return Object
      * @throws Exception
@@ -166,17 +174,14 @@ public abstract class BasicURLClassPath implements ClassPath {
             ClassLoader.registerAsParallelCapable();
         }
 
-        /**
-         *
-         * @param parent ClassLoader
-         */
         public CustomizableURLClassLoader(final ClassLoader parent) {
             super(new URL[0], parent);
         }
 
         /**
-         *
-         * @param url URL
+         * URLs to be added to the URL list for searching classes and resources
+         * 
+         * @param url the URL to be added to the URLs list
          */
         public void add(final URL url) {
             super.addURL(url);
--- a/container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
+++ b/container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
@@ -27,6 +27,11 @@ import java.security.PrivilegedAction;
 import java.util.Locale;
 
 public abstract class BasicURLClassPath implements ClassPath {
+
+    /**
+     *
+     * @return ClassLoader
+     */
     public static ClassLoader getContextClassLoader() {
         return AccessController.doPrivileged(new PrivilegedAction<ClassLoader>() {
             @Override
@@ -39,6 +44,12 @@ public abstract class BasicURLClassPath implements ClassPath {
     private Field ucpField;
     private boolean ucpFieldErrorLogged;
 
+    /**
+     *
+     * @param jar URL
+     * @param loader URLClassLoader
+     * @throws Exception
+     */
     protected void addJarToPath(final URL jar, final URLClassLoader loader) throws Exception {
         final Object cp = getURLClassPath(loader);
         if (cp == null && CustomizableURLClassLoader.class.isInstance(loader)) {
@@ -69,8 +80,8 @@ public abstract class BasicURLClassPath implements ClassPath {
 
     /**
      *
-     * @param dir
-     * @param loader
+     * @param dir File
+     * @param loader URLClassLoader
      * @throws Exception
      */
     protected synchronized void addJarsToPath(final File dir, final URLClassLoader loader) throws Exception {
@@ -119,7 +130,7 @@ public abstract class BasicURLClassPath implements ClassPath {
 
     /**
      *
-     * @param loader
+     * @param loader URLClassLoader
      * @return Object
      * @throws Exception
      */
@@ -159,10 +170,18 @@ public abstract class BasicURLClassPath implements ClassPath {
             ClassLoader.registerAsParallelCapable();
         }
 
+        /**
+         *
+         * @param parent ClassLoader
+         */
         public CustomizableURLClassLoader(final ClassLoader parent) {
             super(new URL[0], parent);
         }
 
+        /**
+         *
+         * @param url URL
+         */
         public void add(final URL url) {
             super.addURL(url);
         }
--- a/container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
+++ b/container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
@@ -132,12 +132,6 @@ public abstract class BasicURLClassPath implements ClassPath {
         }
     }
 
-    /**
-     * 
-     * @param loader URLClassLoader
-     * @return Object
-     * @throws Exception
-     */
     protected Object getURLClassPath(final URLClassLoader loader) throws Exception {
         final Field ucpField = this.getUcpField();
         if (ucpField == null) {
--- a/container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
+++ b/container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
@@ -28,10 +28,6 @@ import java.util.Locale;
 
 public abstract class BasicURLClassPath implements ClassPath {
 
-    /**
-     *
-     * @return ClassLoader
-     */
     public static ClassLoader getContextClassLoader() {
         return AccessController.doPrivileged(new PrivilegedAction<ClassLoader>() {
             @Override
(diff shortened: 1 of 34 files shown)


container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
```diff
@@ -27,6 +27,11 @@
 import java.util.Locale;
 
 public abstract class BasicURLClassPath implements ClassPath {
+
+    /**
```
otaviojava: :question: Does make sense to generate an empty javadoc?

container/openejb-loader/src/main/java/org/apache/openejb/loader/FileUtils.java
```diff
@@ -24,14 +24,20 @@
 @SuppressWarnings("PMD.OverrideBothEqualsAndHashcode")
 public class FileUtils {
 
-    private static final java.util.Random _random = new java.util.Random();
+    private static final java.util.Random RANDOM = new java.util.Random();
```
otaviojava: That is not a static type. to be `RANDOM`
Also, could check if it makes sense to update to [ThreadLocalRandom](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html )

container/openejb-loader/src/main/java/org/apache/openejb/loader/SystemInstance.java
```diff
@@ -263,7 +318,7 @@ public ClassLoader getClassLoader() {
         return removed;
     }
 
-    private static final AtomicReference<SystemInstance> system = new AtomicReference<SystemInstance>();
+    private static final AtomicReference<SystemInstance> SYSTEM = new AtomicReference<SystemInstance>();
```
otaviojava: Could you use a diamond operator?
`private static final AtomicReference<SystemInstance> SYSTEM = new AtomicReference<>();`

cicekhayri: :) missed that changing it now

container/openejb-loader/src/main/java/org/apache/openejb/loader/FileUtils.java
cicekhayri: Okej I will check

container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
cicekhayri: No maybe not :) I remove them

cicekhayri:
@otaviojava I updated little code, could you please check if it's okej

container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
```diff
@@ -39,6 +40,12 @@ public ClassLoader run() {
     private Field ucpField;
     private boolean ucpFieldErrorLogged;
 
+    /**
+     *
```
otaviojava: More empty javadoc :)

container/openejb-loader/src/main/java/org/apache/openejb/loader/FileUtils.java
```diff
@@ -20,18 +20,25 @@
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.Hashtable;
+import java.util.concurrent.ThreadLocalRandom;
 
 @SuppressWarnings("PMD.OverrideBothEqualsAndHashcode")
 public class FileUtils {
 
-    private static final java.util.Random _random = new java.util.Random();
+    private static final ThreadLocalRandom _random = ThreadLocalRandom.current();
```
otaviojava: About the name, sorry my mistake, that is a const, therefore `RANDOM` sounds nice

container/openejb-loader/src/main/java/org/apache/openejb/loader/Files.java
```diff
@@ -245,8 +356,8 @@ public static File mkdirs(final File file) {
     }
 
 
-    // Shutdown hook for recursive delete on tmp directories
-    static final List<String> delete = new ArrayList<String>();
+    // Shutdown hook for recursive DELETE on tmp directories
+    static final List<String> DELETE = new ArrayList<String>();
```
otaviojava: diamond :)

otaviojava: Could you check if this variable can be `private`, I'm not sure

otaviojava:
That is nice @cchacin :) I believe just left the empty javadoc

container/openejb-loader/src/main/java/org/apache/openejb/loader/FileUtils.java
cicekhayri: :)

container/openejb-loader/src/main/java/org/apache/openejb/loader/Files.java
cicekhayri: absolutely

cicekhayri:
found few more diamond :)

container/openejb-loader/src/main/java/org/apache/openejb/loader/FileUtils.java
dblevins: Not this isn't safe usage of `ThreadLocalRandom`.  Whatever thread initializes `FileUtils` will have its not-thread-safe instance of ThreadLocalRandom saved in a static field and shared thereafter by all threads.  In short, this code grabs what is intended to be thread-scoped and stores it into the java heap which is seen/shared by all threads.  You need to call `ThreadLocalRandom.current()` in each method that needs a `ThreadLocalRandom`.

cicekhayri: Okej, is it best to revert it back for now

otaviojava: hey, what David means is instead of having the field, just use it every time.

E.g:

```java
public class FileUtils {

//not more the static field

    private File home;

    private FileUtils(final String homeDir, final String defaultDir) {
        this(homeDir, defaultDir, SystemInstance.get().getProperties());
    }

 public static File createTempDirectory(final String pathPrefix) throws IOException {
        for (int maxAttempts = 100; maxAttempts > 0; --maxAttempts) {

            final String path = pathPrefix + ThreadLocalRandom.current().nextLong();
            final File tmpDir = new File(path);

            if (!tmpDir.exists() && tmpDir.mkdirs()) {
                return tmpDir;
            }
        }
        throw new IOException("Cannot create temporary directory at: " + pathPrefix);
    }
```

cicekhayri:
@otaviojava could you please check if it's right 🤞

otaviojava:
@cicekhayri that is fine, please, just check the Javadoc

cicekhayri:
I'll check them and try to fix them

cicekhayri:
Hi everyone,  wonder if someone could check the JavaDocs I added..would love some help if there are wrong ones

container/openejb-loader/src/main/java/org/apache/openejb/observer/event/ObserverRemoved.java
```diff
@@ -26,10 +26,20 @@
 
     private final Object observer;
 
+    /**
+     * Remove the observable object so this object doesn't have any observers
```
pzygielo: I don't think that's true. It is constructor, and it apparently does not remove anything.

container/openejb-loader/src/main/java/org/apache/openejb/observer/event/ObserverAdded.java
```diff
@@ -26,10 +26,20 @@
 
     private final Object observer;
 
+    /**
+     * Add the provided object to be observable
```
pzygielo: That is not what happens here.

container/openejb-loader/src/main/java/org/apache/openejb/observer/event/ObserverAdded.java
```diff
@@ -26,10 +26,20 @@
...
+     * 
+     * @param observer
+     */
     public ObserverAdded(final Object observer) {
         this.observer = observer;
     }
 
+    /**
+     * Return the observable object
```
pzygielo: Do you consider observable the same as observer?

container/openejb-loader/src/main/java/org/apache/openejb/observer/ObserverManager.java
```diff
@@ -74,6 +80,12 @@ public boolean addObserver(final Object observer) {
         }
     }
 
+    /**
+     * Remove the observable object so this object doesn't have any observers
```
pzygielo: I think only specified `observer` will be removed, and all the rest will be kept.
And observable will not be removed.

container/openejb-loader/src/main/java/org/apache/openejb/loader/FileUtils.java
```diff
@@ -120,10 +150,19 @@ public File getFile(final String path, final boolean validate) throws IOExceptio
         return file;
     }
 
+    /**
+     * Creates Temp Directory
+     * 
+     * Throws IOException if directory doesn't exists and if it fails to create directories
```
pzygielo: - It belongs to `@throws` tag, I think.
- > if directory doesn't exists 
1. exist~~s~~
1. Which one? The one we want to create? We have no control over random used in the method, so how can programmer react to that IOEx?
- > doesn't exists and if it fails

What happens if all 100 *candidate directories* exist already? I see the method will throw as well. So your description is not quite correct.

cicekhayri:
This is bad 😓, i think i just delete this PR and let somebody else work on this

cicekhayri:
Hej everyone, got some comments on the JavaDocs I did and It's sounds that it's really bad and wonder if you guys also thinks that.. and wonder if I just delete the JavaDocs and just commit the diamond op. or should I just delete whole branch and start from scratch.. I need to get my motivation up again :) ..

pzygielo:
:+1: for `<>` changes

container/openejb-loader/src/main/java/org/apache/openejb/loader/BasicURLClassPath.java
```diff
@@ -111,6 +132,12 @@ public boolean accept(final File dir, String name) {
         }
     }
 
+    /**
+     *
```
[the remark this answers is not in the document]
cicekhayri: I removed the empty javadocs

pzygielo:
>  otaviojava approved these changes 7 minutes ago 

:disappointed:

pzygielo:
@otaviojava oh, I see. I thought it was full green light for merge.

otaviojava:
@pzygielo 
Once three persons are reviewing that PR that means at least two persons need to set green. The best scenario will be if all of then approve.

container/openejb-loader/src/main/java/org/apache/openejb/observer/event/ObserverRemoved.java
jeanouii: This is actually the event thrown when an observer is removed. All remaining observers have the ability to observe this event and react if needed.

container/openejb-loader/src/main/java/org/apache/openejb/observer/event/ObserverAdded.java
jeanouii: Same as previously but for adding a new observer in the system. This is also an event

jeanouii:
I am overall ok with the PR but the javadoc part is more the one to discuss.

I agree with @pzygielo. If JavaDoc does not bring anything more than you can read with the code itself, then I just remove it.

JavaDoc listing the attributes and the throws are useless to me because I can read it on the code itself. Remember that the best documentation is the code. So if the code is not meaningful, the javadoc won't make it clear. So I'm tempted to address the root cause first and fix the code.

Than being said, @cicekhayri can you look at @pzygielo comments and take them into account so I can merge this PR before it grows and conflicts start appearing?

cicekhayri:
@jeanouii I removed the JavaDocs because it doesn't add anything useful to the code

jeanouii:
Sounds good, let's get the PR merged then.
Thanks @cicekhayri for the PR and thanks @pzygielo for the review.

PR #275, opened 2018-12-14
TOMEE-2357 - Fixed issue and updated test on example application.

cesarhernandezgt:


mp-jwt/src/main/java/org/apache/tomee/microprofile/jwt/jaxrs/MPJWTSecurityAnnotationsInterceptor.java
```diff
@@ -39,14 +39,19 @@ public void filter(final ContainerRequestContext requestContext) throws IOExcept
         }
 
         final Set<String> roles = rolesAllowed.get(resourceInfo.getResourceMethod());
+
         if (roles != null && !roles.isEmpty()) {
             final SecurityContext securityContext = requestContext.getSecurityContext();
+            Boolean hasAtLeasOneValidRole = false;
```
pzygielo: Why not use primitive `boolean`? Avoid boxing here and in L48 and unboxing in L52.

cesarhernandezgt: Nice catch @pzygielo !
fixed.

jeanouii:
Thanks @pzygielo for the review.
All fixed. I tested locally and looks good.

I'll merge and monitor the build to see if it's ok.

PR #276, opened 2018-12-14
TOMEE-2363 Introduces OWASP dependency check via profile

rzo1:
This PR introduces the [OWASP Dependency Check Maven Plugin](https://jeremylong.github.io/DependencyCheck/index.html) in a basic configuration at the parent pom level of the project.

As the dependency checking is quite expensive (see time below), I added it is a separate profile `owasp`. To enable it in the Maven build process, you just need to add `-Powasp-report` or `-Powasp-check`. 
An aggregated report `dependency-check-report.html` is created in the target directory of the root project (only in acse you enabled `-Powasp-report`)

I also added some exclusions related to false positives (see `owasp-dc-suppression.xml`).
I also added the aggregated output for the run on my second system: [dependency-check-report.zip](https://github.com/apache/tomee/files/2690169/dependency-check-report.zip)

Some timings for my system (here Windows, on Linux it is a  lot faster...)

QuickBuild without OWASP: 07:08 min
QuickBuild with OWASP 1st RUN: 26:02 min (pre-caching vulnerabilities)
QuickBuild with OWASP 2nd+ RUN: 13:06 min

Note, with `-Powasp-check` the  the build will fail for a CVE score >= 8.0. However, we should decide on a common CVE score value to do so on the CI systems. 

f this PR holds your expectations, I can backport this to 7.0.x and 7.1.x branches. In a next step, we can analyze the outcomes and create JIRA issue and/or enhance the supress configuration for false positives.

### DIFF
--- a/owasp-dc-suppression.xml
+++ b/owasp-dc-suppression.xml
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.2.xsd">
+    <suppress>
+        <notes><![CDATA[
+   file name: self dependencies...
+   ]]></notes>
+        <gav regex="true">^org\.apache\.tomee:.*$</gav>
+        <cve>CVE-2018-8031</cve>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[
+   file name: self dependencies...
+   ]]></notes>
+        <gav regex="true">^org\.apache\.tomee:.*$</gav>
+        <cve>CVE-2010-1151</cve>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[
+   file name: false positive apache http server
+   ]]></notes>
+        <gav regex="true">^org\.apache\.tomee:.*$</gav>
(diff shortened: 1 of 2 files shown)


jeanouii:
HI,

Thanks for the PR and the details.
From my experience, in this kind of situation, if the check is not activated and does not make the build to fail, it's not so useful. 

So I'd VOTE to at least have the profile activated on our CI system (buildbot). And if it fails it needs to fail the build.

Thoughts?

jeanouii:
BTW, happy to check that in as a first step and you can create another PR for iteration 2.

rzo1:
Hey  @jeanouii & @radcortez 

thanks for the feedback - I can update this PR.

I would propose:

- (1) I will add a profile `owasp-report` to generate the aggregated HTML report, which can be triggered manually as it takes a long time
- (2) I will add a profile `owasp-check`, which does not generate HTML reports (which will speed up this process) and which will **fail** the build for CVE's greater a specific value. This profile can then be added on the CI System (build bot), which may support caching the owasp CVE files (so we will get a speed boost).

What do you think?

radcortez:
@rzo1 that sounds great. I'll try it.

PR #289, opened 2018-12-17
TOMEE-2301  Concurrency utilities example

brunobat:
This is the second part of the example. Adding a thread factory.
Will improve the example by using CountDownLatch.


brunobat:
Hey, This should be ok for reviewing now.

PR #301, opened 2018-12-21
Fixes TOMEE-2261 Make service.bat compatible with JDK11+

gerwinjansen:
When JAVA_HOME is set to a path with a space (e.g. C:\Program Files\Java\jdk11.0.1), I get this error:
```
C:\Program Files\TomEE\bin>service install
11 was unexpected at this time.
```
This is introduced with #178. I fixed it by adding quotes to the path.

Bonus: removal of superfluous tokens which are not used.


jgallimore:
I've merged this in, many thanks for the PR!

PR #316, opened 2018-12-27
Interceptor titles

pavelsmajda:
It seems that title was ignored when there is another title under the top section. I removed so it should be displaying titles correctly now.

Issue https://issues.apache.org/jira/browse/TOMEE-2411

### DIFF
--- a/examples/cdi-alternative-and-stereotypes/README.md
+++ b/examples/cdi-alternative-and-stereotypes/README.md
@@ -3,7 +3,6 @@ type=page
 status=published
 title=CDI Alternative and Stereotypes
 ~~~~~~
-# Introduction
 
 CDI is a revolution for the Java EE world. This specification is the best one to avoid coupling between classes.
 
--- a/examples/simple-cdi-interceptor/README.md
+++ b/examples/simple-cdi-interceptor/README.md
@@ -3,7 +3,6 @@ type=page
 status=published
 title=Simple CDI Interceptor
 ~~~~~~
-#Simple CDI Interceptor
 
 Let's write a simple application that would allow us to book tickets for a movie show. As with all applications, logging is one cross-cutting concern that we have. 
 


josehenriqueventura:
Thank you for your PR.
I tried to generate the .html file based on those two .md files that you modified, but they were generated with the title after the following commit: https://github.com/apache/tomee/commit/0efe6d0b9ca044d2be312f7baa4b0918a059ee6c

The **tomee-site-generator** searches for the String "title:" instead of "#...". 
https://github.com/apache/tomee-site-generator/blob/ee0724fcdf3543b014a3869ebbd22512159499ba/src/main/java/org/apache/tomee/website/JbakeHeaders.java#L118

If you haven't yet, you might try to checkout tomee-site-generator project and try out adding the title to file and see the results.
Checkout  https://github.com/apache/tomee-site-generator/
- mvn clean compile
- tomee-site-generator/target/site-1.0-SNAPSHOT/tomee-8.0/examples/cdi-alternative-and-stereotypes.html (double click and the title is NULL)
- then you can modify this file **/tomee-site-generator/repos/tomee-8.0/examples/cdi-alternative-and-stereotypes/README.md**, just adding the title as was added in this commit https://github.com/apache/tomee/commit/0efe6d0b9ca044d2be312f7baa4b0918a059ee6c;
- compile the project again
- **tomee-site-generator/target/site-1.0-SNAPSHOT/tomee-8.0/examples/cdi-alternative-and-stereotypes.html** (double click and the title should be displayed).

I hope I have helped. =)

josehenriqueventura:
@pavelsmajda 
They were generated correctly now because that commit was merged into the master branch and then you cloned the **tomee-site-generator**, it checks out the latest revision of the repository, and the last one already contains that fix.

If you removed the "title:.." from the README.md file, you probably will get that NULL and 
I got that error here because I had a cached version of master repo which didn't have that fix. =)

So.. I think this issue is already fixed.

> Also, I saw a task and email thread about migration to adoc format, should I migrate docs do .adoc instead of .MD files?
I'm not sure how the guys are processing with these changes. I would say that would be good to ask it in the dev list. =)

brunobat:
@pavelsmajda go you need some help with the rest of the PR?

PR #323, opened 2018-12-28
Optmizes try catch

otaviojava:
* Optimizes the try using the try-with-resources that allows us to declare resources to be used in a try block with the assurance that the resources will be closed when after execution of that block.  An ARM automatic resource management by the JVM.

*  The use of multi-catch that removes redundancies catch treatment.


otaviojava:
Thank you for the review.
Can someone merge this?

PR #326, opened 2018-12-28
tomee-directory-structure.html formatting additions

pavelsmajda:
- Added title for tomee-directory-structure.html file
- Decreased font size for TomEE Directory Layout
- Changing *$tomee-install-dir>* to be in strong text format

### DIFF
--- a/docs/tomee-directory-structure.adoc
+++ b/docs/tomee-directory-structure.adoc
@@ -2,7 +2,7 @@
 :jbake-date: 2018-12-05
 :jbake-type: page
 :jbake-status: published
-
+:jbake-title: TomEE Directory Structure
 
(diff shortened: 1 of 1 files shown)


docs/tomee-directory-structure.adoc
```diff
@@ -17,9 +17,9 @@ CONDITIONS OF ANY KIND, either express or implied. See the License for
...
 License.
 
-= TomEE Directory Layout:
+== TomEE Directory Layout:
 
 TomEE directory layout is the same as that of Tomcat, with a few changes
 as described below.
 
-Considering this root to be the $tomee-install-dir>
+Considering this root to be the *$tomee-install-dir>*
```
cesarhernandezgt: we can use `` instead of ** to make it code formating.

PR #333, opened 2018-12-31
TOMEE-2333 MicroProfile JWT Example for mp.jwt.verify.publickey configuration

ivannov:


ivannov:
Nice catch @brunobat :smile:

PR #335, opened 2018-12-31
Fixed a few issues related to closing of resources

martinfmi:
Fixed a few issues related to closing of resources uncovered from SonarQube

Signed-off-by: Martin Toshev <marto8808@gmail.com>


martinfmi:
Thank you, Bruno. I've reverted the change as the writer should not be closed in the finally block.

PR #338, opened 2018-12-31
Use correct line separator when checking call result.

doychin:
Fix for issue 
https://issues.apache.org/jira/browse/TOMEE-2437


jgallimore:
That's awesome, thank you!

PR #340, opened 2019-01-01
TOMEE-2289 microprofile openAPI example

puneethps:
The JIRA status for TOMEE-2289 hasn't changed for quite sometime so I felt I could add a new project for this. Can someone please  review the PR and suggest changes required?


examples/mp-openapi/src/main/java/org/superbiz/openapi/WeatherService.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Path("/status/{day}/")   
+    @GET
+    @Operation(summary = "Finds weather for day specified in the URL ", description = "Describes how the day will be.")
+    @APIResponse(responseCode = "400", description = "Weather for this day not found")
+    public Response dayStatus( @Parameter(description = "The day for which the weather needs to be fetched.", required = true) @PathParam("day") String day	) {
+    	
+    	if(day.equalsIgnoreCase("tomorrow"))
+    		return Response.status(Response.Status.NOT_FOUND).build();
+    	String message=day+ "is a sunny day.";
```
ivanjunckes: It is missing a space. Output is "Mondayis a sunny day." and should be "Monday is a sunny day."

examples/mp-openapi/README.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+== Response
+....
+{today} is a sunny day.
+
+
+== OpenAPI
+
+OpenAPI is a spec for natively producing OpenAPI v3 documents from JAX-RS applications. 
+There exists an endpoint /openApi which gives the OpenAPI document based on Accept request header.
```
ivanjunckes: "/openApi" is not a valid path, it must be "/openapi". Path is casesensitive.

examples/mp-openapi/README.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+
+== Request
+....
+GET http://localhost:8080/mp-openapi/openapi
```
ivanjunckes: It does not work without content-type as application/json so it is worth mention it. This is a bug in openapi I think. Because it should have a default message body reader/writer and it should be yaml. But it can be fixed later, there is a discussion on the list about it.

examples/mp-openapi/src/main/java/org/superbiz/openapi/WeatherService.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.core.Response;
+
+import org.eclipse.microprofile.openapi.annotations.Operation;
+import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
+import org.eclipse.microprofile.openapi.annotations.media.Content;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
```
ivanjunckes: Can you be careful formatting the class? There are a lot of unnecessary spaces.

examples/mp-openapi/src/main/java/org/superbiz/openapi/Weather.java
```diff
@@ -0,0 +1,25 @@
+package org.superbiz.openapi;
+
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@Schema(name="Weather", description="POJO that represents weather.")
```
ivanjunckes: Can you follow a pattern with spaces? name="Wather" has no space between name and the value. But for @Schema the variables have space. Can you make sure to format them correctly having one space between the variable and the value?

examples/mp-openapi/src/main/java/org/superbiz/openapi/WeatherService.java
```diff
@@ -0,0 +1,52 @@
...
+        	      .entity(message)
+        	      .build();        	         	
+    }
+
+    @Path("/detailedWeather/{day}/")   
+    @GET
+    @APIResponse(description = "Detailed Weather", content=@Content(mediaType="application/json", schema= @Schema(implementation=Weather.class)))
+    public Weather getDetailedWeather( @Parameter(description = "The day for which the detailed weather needs to be fetched.", required = true) @PathParam("day") String day) {
+    	
+        Weather w=new Weather();
```
ivanjunckes: Space is missing in several places, between variable and value assingments. Please format the class correctly.

examples/mp-openapi/src/test/java/org/superbiz/openapi/WeatherServiceTest.java
```diff
@@ -0,0 +1,208 @@
...
+                .get();
+        final String metaData = message.readEntity(String.class);
+        JsonObject jsonObject = Json.createReader(new StringReader(metaData)).readObject();
+        JsonObject response=jsonObject.get("paths").asJsonObject().get("/weather/detailedWeather/{day}").asJsonObject();
+        final String expected = "{\r\n" + 
+        		"            \"get\": {\r\n" + 
+        		"                \"operationId\": \"getDetailedWeather\",\r\n" + 
+        		"                \"parameters\": [\r\n" + 
+        		"                    {\r\n" + 
+        		"                        \"allowEmptyValue\": false,\r\n" +
```
ivanjunckes: There are a lot of values in the response that are not in the Weather pojo. This seems to me like a bug in the implementation. Can you evaluate and raise this bug on the list?
allowEmptyValue
allowReserved
required
... are not part of the pojo.

examples/mp-openapi/README.adoc
puneethps: I think it is working as expected now, without the Accept:application/json header it is outputting a valid YAML. I think I should document that in the README file.

puneethps:
I noticed one more difference from the spec and the implementation, what I understood from the spec was that there is supposed to be a components field in the yaml or the json if there is a schema annotation which I don't see in the generated output. I think this should also be raised as an issue.

examples/mp-openapi/README.adoc
ivanjunckes: I still get "No message body writer has been found for class org.apache.geronimo.microprofile.openapi.impl.model.OpenAPIImpl, ContentType: text/html" without the accept header.

ivanjunckes: It is a bug, so you just need to add the header to the documentation. And if possible add a ticket in the Geronimo JIRA.

examples/mp-openapi/src/test/java/org/superbiz/openapi/WeatherServiceTest.java
```diff
@@ -0,0 +1,203 @@
...
+    public void detailedDayStatus() {
+        WebTarget webTarget = this.client.target(this.base.toExternalForm());
+        final Response message = webTarget.path("/openapi")
+                .request()
+                .accept(MediaType.APPLICATION_JSON)
+                .get();
+        final String metaData = message.readEntity(String.class);
+        JsonObject jsonObject = Json.createReader(new StringReader(metaData)).readObject();
+        JsonObject response = jsonObject.get("paths").asJsonObject().get("/weather/detailedWeather/{day}").asJsonObject();
+        final String expected = "{\r\n" +
```
ivanjunckes: This json doesn't look well too, but this is an issue that needs to be reported as well. Like I said before there are properties that are no in the Weather class.

examples/mp-openapi/README.adoc
puneethps: I think the browser by default sends ContentType: text/html as header if I do the same request in Postman I get a YAML, so I dont think a ticket is required.

examples/mp-openapi/src/test/java/org/superbiz/openapi/WeatherServiceTest.java
puneethps: I'll raise an issue regarding this.

examples/mp-openapi/README.adoc
ivanjunckes: Ok I opened a PR in geronimo to tackle this https://github.com/apache/geronimo-openapi/pull/4.

PR #342, opened 2019-01-02
TOMEE-2332 MP-jwt-jwk example

cotnic:
Implemented the MP-jwt for JWKs public key example.

Also included the usage of MP-rest-client for testing.


examples/mp-rest-jwt-jwk/README.adoc
```diff
@@ -0,0 +1,76 @@
...
+<dependency>
+    <groupId>org.eclipse.microprofile.jwt</groupId>
+    <artifactId>microprofile-jwt-auth-api</artifactId>
+    <version>${mp-jwt.version}</version>
+    <scope>provided</scope>
+</dependency>
+....
+. Annotate our `Application.class` with `@LoginConfig(authMethod = "MP-JWT")`
+
+. Provide public and private key for authentication. And specify the location of the public key and the issuer in our
```
jeanouii: What's actually required is the public key because MicroProfile JWT targets the validation side (consuming a JWT). The private key would be useful on the producer side of things such as an identity provider, or an API Gateway.

examples/mp-rest-jwt-jwk/README.adoc
```diff
@@ -0,0 +1,76 @@
...
+    "guest", "admin"
+  ]
+}
+----
+
+== Access the endpoints with JWT token
+
+We access endpoints from our test class by creating a `JWT` with the help of
+our `TokenUtils.generateJWTString(String jsonResource, String keyId)` which signs our user
+data in json format with the help of our `src/test/resources/{keyId}` key.
```
jeanouii: This is where you need the private key, but this is for testing purpose to generate a valide and signed JWT

examples/mp-rest-jwt-jwk/pom.xml
```diff
@@ -0,0 +1,215 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+            <scope>test</scope>
+        </dependency>
+
+        <!--
```
jeanouii: Small detail, but looks like some test dependencies are before this section, not sure it's intended or not. 
It's a detail so won't prevent the merge at all

cotnic:
@jeanouii I've changed and improved the README and pom.xml based on your comments.

radcortez:
Looks great @cotnic . Merged!

PR #343, opened 2019-01-02
(TOMEE-2438) Centralise TomEE and OpenEJB system properties

gerdogdu:
Initial attempt for TomEE system config.

### DIFF
--- a/tomee/pom.xml
+++ b/tomee/pom.xml
@@ -34,6 +34,7 @@
   <name>OpenEJB :: TomEE</name>
 
   <modules>
+    <module>tomee-config</module>
     <module>tomee-loader</module>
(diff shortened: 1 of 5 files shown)


gerdogdu:
Any update on this?

jgallimore:
Hi Gurkan, I'll take a look this afternoon for you.

jgallimore:
Is there a rationale behind creating a whole new module for this? I'm not against the centralization, but I would imagine this could live in OpenEJB core. What's the long-term goal for this?

What would be particularly great is if this could feed into the website somehow - perhaps it could be picked up by David's site generation code so all the properties, along with their documentation end up on a page there automatically. What do you think?

PR #344, opened 2019-01-02
On Windows getJarLocation returns path like x:\tomee\bla\bla\bla

doychin:
This is not ok for the code that uses this path and tries to convert it to URL.

For that reason a valid URL string must be returned from this method.

Also the cycle below cycles on ejb modules but generates result from appModule which is already covered before that. Am I wrong here on my assumption?

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/config/CmpJpaConversion.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/config/CmpJpaConversion.java
@@ -267,10 +267,12 @@ public class CmpJpaConversion implements DynamicDeployer {
 
     private String getPersistenceModuleId(final AppModule appModule) {
         if (appModule.getModuleId() != null) {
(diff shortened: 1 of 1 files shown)


jgallimore:
@doychin is there a test that fails without this change? It looks great - just wanted to check it here.

doychin:
@jgallimore Yes, all tests for persistence in openejb-core.
The problem is that the path produced from old code is Windows path that is not directly compatible with URL format. And that path is passed to PersistenceModule constructor as rootUrl. So all tests related to CmpJpaConversion fail with exception

java.lang.IllegalArgumentException: Illegal character in opaque part at index 2: D:\tomee\container\openejb-core\target\TestModule

doychin:
Here is a list of tests that fail in openejb-core only: OpenEjb2ConversionTest, SunCmpConversionTest, SimpleCmpTest, StatefulJPATest, JPACMDIntegrationTest

PR #349, opened 2019-01-02
Some of the returned values are performance dependant - stddev

doychin:
This one is more complex as change. stdDev value in histogram depends on how fast values are added to the histogram on the server. In order to reuse existing expected values and filter the value that is not always the same, I've had to implement some parsing before comparing the returned string with expected string.
In case you don't like it I'm open for suggestions.

### DIFF
--- a/examples/mp-metrics-histogram/src/test/java/org/superbiz/histogram/WeatherServiceTest.java
+++ b/examples/mp-metrics-histogram/src/test/java/org/superbiz/histogram/WeatherServiceTest.java
@@ -31,6 +31,7 @@ import org.junit.runner.RunWith;
 import javax.json.Json;
 import javax.json.JsonObject;
 
+import javax.json.stream.JsonParser;
 import javax.ws.rs.client.Client;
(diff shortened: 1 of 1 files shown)


brunobat:
Hi @doychin. 
Thanks for the PR.
Can you please resolve the conflict?

doychin:
Conflict was introduced by commit https://github.com/apache/tomee/commit/e7e80823b6014eb2e2a45d25234a5b661eeed8b1

This commit also contains the fix I propose in #347.

brunobat:
Ok. If the fix is already in, probably you can close this PR.

doychin:
No. Only the fix for PR #347 is in. The problem with that commit is that it changes 2 files. One that introduces the conflict and that change is not OK. The test will fail again on next run.

The other is the same fix I propose in #347. 

If you can revert that commit then you should be able to apply both PR #347 and this (#349)

radcortez:
I've rebased and fix the conflict. Should be ok now. Thank you @doychin

PR #352, opened 2019-01-03
TOMEE-2443 - Some code improvements on the Mp JWT module

ivanjunckes:


mp-jwt/src/main/java/org/apache/tomee/microprofile/jwt/principal/JWTCallerPrincipalFactory.java
```diff
@@ -17,20 +17,22 @@
...
+import java.util.logging.Logger;
 
 /**
  * The factory class that provides the token string to JWTCallerPrincipal parsing for a given implementation.
  */
 public abstract class JWTCallerPrincipalFactory {
 
     private static JWTCallerPrincipalFactory instance;
-
+    private static final Logger logger = Logger.getLogger(MPJWTCDIExtension.class.getName());
```
radcortez: Should the Logger be the CDI Extension class?

mp-jwt/src/main/java/org/apache/tomee/microprofile/jwt/config/JWTAuthContextInfo.java
```diff
@@ -16,19 +16,22 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 /**
  * The public key and expected issuer needed to validate a token.
  */
 public class JWTAuthContextInfo {
+    private static final Logger logger = Logger.getLogger(MPJWTCDIExtension.class.getName());
```
radcortez: Should the Logger be the CDI Extension class?

mp-jwt/src/main/java/org/apache/tomee/microprofile/jwt/config/ConfigurableJWTAuthContextInfo.java
```diff
@@ -67,6 +67,8 @@
 
     private Config config;
     private JWTAuthContextInfo jwtAuthContextInfo;
+    private static final String PUBLIC_KEY_ERROR = "Could not read MicroProfile Public Key";
+    private static final String PUBLIC_KEY_ERROR_LOCATION = PUBLIC_KEY_ERROR + " from Location: ";
```
radcortez: I'm not sure if we should follow this pattern for message logging, considering that we don't do it for the entire codebase.

ivanjunckes: I just didn't want the code to repeat the message a lot of times. What is the issue with this pattern?

ivanjunckes: I think it is just an improvement on this class, but I think this is a good practice instead of replicating the same thing everywhere. If I find other places like this I will change.

Do you think this is a blocker for this PR? If not can you approve?

radcortez: No, not a blocker. Should be fine.

radcortez:
@ivanjunckes actually, can you remove this commit: 
https://github.com/apache/tomee/pull/352/commits/d7b88002eef05e41286dc8a3c133d10182e28060?

Doesn't seem to fit with the PR scope.

PR #359, opened 2019-01-06
TOMEE-2390: Added README.adoc for the multiple-tomee-arquillian example.

cotnic:


brunobat:
Hi @cotnic thanks for the documentation.
Would it be possible to also mention the sever configuration on the arquillian.xml file?
That is required for the example to work properly.

PR #360, opened 2019-01-07
Cleanup test code and also minor changes in the tested class.

doychin:
- I made the stack trace produced by auto connection tracker look better
- some generics parameters and code cleanup.
- moved some duplicate code to separate method.

Probably we will need to discuss the details of the test. It will be good if I can check the original bug report to see the reasons behind the initial test implementation.


container/openejb-core/src/test/java/org/apache/openejb/resource/AutoConnectionTrackerTest.java
```diff
@@ -154,38 +152,29 @@ public void test() throws Exception {
...
-                    }
+            runTest(() -> {
+                try {
+                    bean.nonleakyNonTxMethod();
+                } catch (Exception e) {
+                    e.printStackTrace();
                 }
             });
 
-            System.gc();
```
brunobat: same here

container/openejb-core/src/test/java/org/apache/openejb/resource/AutoConnectionTrackerTest.java
```diff
@@ -154,38 +152,29 @@ public void test() throws Exception {
 
         {
             logCapture.clear();
-            runTest(new Runnable() {
-                @Override
-                public void run() {
-                    try {
-                        bean.nonLeakyTxMethod();
-                        System.gc();
```
brunobat: I suspect that System.gc() is in there on purpose to simulate something odd. Please don't remove that.

container/openejb-core/src/test/java/org/apache/openejb/resource/AutoConnectionTrackerTest.java
```diff
@@ -214,30 +203,7 @@ public void run() {
 
     // this is a very quick and dirty hack for debugging purpose
     private void assertLogs(final LogCaptureHandler logCapture, final int times, final String message) {
-        final int iteration = 5;
```
brunobat: This might take some time to return, hence the wait... And I suspect the GC is related to that.

doychin: Actually that wait is not needed here. All logs are captured by LogCaptureHandler. All logs are added synchronously to the handler while calls to business code are executed. 

Asynchronous test code actually waits for all business calls to return so when assertLogs is called to check for specific messages all messages are logged there and there is no need to wait. 

For synchronous calls again there is no need to wait. All code is executed in the main thread.

container/openejb-core/src/test/java/org/apache/openejb/resource/AutoConnectionTrackerTest.java
doychin: I restored back System.gc(); in the new commit.

brunobat:
@doychin I've run the code and it looks ok to me, but I don't know if the changes are subverting the original purpose of the test.

doychin:
Actually test was failing only due to the  speed of placing leaked connection reference in the reference queue is not constant and at some point especially during heavy GC the time needed to put these leaked references increases. If calls to business methods are to fast like in the case with the test - all are local calls JVM executes the call code fast enough to get to setEnvirment method before references reached the queue. 

I think the whole problem comes from the fact that this code relies on the GC. this is a big black box and you never know how it will behave in different environments.  

I considered complete rewrite of this code and moving these checks after business method execution in an invocation handler but I'm not very familiar with that part of tomee. 

I think a handler that will put some marker before method invocation and check for allocated connections for that marker is better and more predictable then relaying on GC to put allocated instances and make the check if and when there is a new call for connection allocation.

doychin:
Is it safe to recommend to do setEnvironment code asynchronous to the calling thread? This will introduce some latency before the call to pull reference from phantom queue. Code will not hurt call performance and only drawback is that log processing in the test should wait some time before declaring success or failure

brunobat:
I'd ask @jgallimore to give his opinion on this.

doychin:
Sorry for the mess ;-(

doychin:
I've implemented the idea to run setEnvironment in separate thread. Previous implementation had the problem that it will introduce a delay if there are active references. 

Now this implementation does not need to check for anything and just checks 1ms later about phantom reference in another thread.

container/openejb-core/src/main/java/org/apache/openejb/resource/AutoConnectionTracker.java
```diff
@@ -58,16 +60,18 @@
     private final TransactionManager txMgr;
     private final Logger logger = Logger.getInstance(LogCategory.OPENEJB_CONNECTOR, "org.apache.openejb.resource");
     private final ConcurrentMap<ManagedConnectionInfo, ProxyPhantomReference> references = new ConcurrentHashMap<>();
-    private final ReferenceQueue referenceQueue = new ReferenceQueue();
+    private final ReferenceQueue<ConnectionInvocationHandler> referenceQueue = new ReferenceQueue<>();
     private final ConcurrentMap<Class<?>, Class<?>> proxies = new ConcurrentHashMap<>();
     private final ConcurrentMap<Class<?>, Class<?>[]> interfaces = new ConcurrentHashMap<>();
 
     private final boolean cleanupLeakedConnections;
+    private final ExecutorService executorService;
```
AndyGee: When using a service, the service should be shut down. I'm not sure right now where the best place for that is, but I'd still like to see the call somewhere.

doychin: There is new commit that handles this.

AndyGee:
@jgallimore this looks good to me. WDYT?

PR #365, opened 2019-01-08
Add Safeguard v1.2

brunobat:
Also fixed a bug in the fallback example. Fallback handlers need to be CDI beans. 
There are still tests failing. See mailing list for details.


jeanouii:
It looks good so let’s get it in

PR #369, opened 2019-01-10
added Docker to mp metrics counted example

carljmosca:
This adds Docker to the example.  It does not require the root user and the application is installed as the ROOT application.  It also corrects one of the URLs .


carljmosca:
Cool, picked something easy for my first. :)

PR #370, opened 2019-01-10
Updates the code using new API facilities

otaviojava:
## Updates the API

This PR has the goal to update the code using the newest API until Java 8. These improvements are:

* Use switch instead of multiple if/else: The switch is faster because there are special bytecodes that allow efficient switch statement evaluation when there are a lot of cases.

* Simplifies code using Objects.equals: It reduces the number of lines of code using the [Objects.equals ](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-)also it will be faster considering the JIT factor.

* Simplifies the primitive compare using [Integer.compare](https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#compare-int-int-)

* Reduces code using removeif condition in the collection.

* from the JVM specification it repleaces [`L` instead of `l`](https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.10).

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/Injection.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/Injection.java
@@ -17,6 +17,8 @@
 
 package org.apache.openejb;
 
+import java.util.Objects;
+
 /**
  * @version $Rev$ $Date$
  */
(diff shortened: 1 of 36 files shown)


radcortez:
Not sure if we should change the .equals() implementation.

This is mostly generated code, so the changes will be lost if we need to regenerate the equals / hashcode. I believe most people just delete the current methods and generate them again.

brunobat:
@otaviojava Any feedback ok @radcortez comments?

otaviojava:
Hey Bruno, I like the idea to standardize to use the new API. It simplifies and Eclipse also can generate using the new Java 7 API.

otaviojava:
That is fine to move it forward?

PR #371, opened 2019-01-11
Startup and shutdown fixes - TOMEE-2454

doychin:
Properly clear all registered thread context listeners that are not initialized as static instances.
Run all tests locally on Linux and Windows.
All pass.

Check https://issues.apache.org/jira/projects/TOMEE/issues/TOMEE-2454 for details.

- next is to re-enable EnvEntryTest from arquillian-tomee-jms-tests and see if that will break build on build bot.

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/cdi/CdiBuilder.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/cdi/CdiBuilder.java
@@ -37,12 +37,14 @@ public class CdiBuilder {
     private static final Logger logger = Logger.getInstance(LogCategory.OPENEJB_CDI, CdiBuilder.class);
     private static final ThreadSingletonService SINGLETON_SERVICE = new ThreadSingletonServiceImpl();
 
+    private static final RequestScopedThreadContextListener REQUEST_SCOPED_THREAD_CONTEXT_LISTENER = new RequestScopedThreadContextListener();
+
     public void build(final AppInfo appInfo, final AppContext appContext, final List<BeanContext> allDeployments) {
         initSingleton().initialize(new StartupObject(appContext, appInfo, allDeployments));
     }
 
     private ThreadSingletonService initSingleton() {
-        ThreadContext.addThreadContextListener(new RequestScopedThreadContextListener());
+        ThreadContext.addThreadContextListener(REQUEST_SCOPED_THREAD_CONTEXT_LISTENER);
         ThreadSingletonService singletonService = SystemInstance.get().getComponent(ThreadSingletonService.class);
 
         //TODO hack for tests.  Currently initialized in OpenEJB line 90.  cf alternative in AccessTimeoutTest which would
@@ -62,6 +64,7 @@ public class CdiBuilder {
 
     public static synchronized ThreadSingletonService initializeOWB() {
         logger.info("Created new singletonService " + SINGLETON_SERVICE);
+        SystemInstance.get().addObserver(SINGLETON_SERVICE);
         SystemInstance.get().setComponent(ThreadSingletonService.class, SINGLETON_SERVICE);
         try {
             WebBeansFinder.setSingletonService(SINGLETON_SERVICE);
@@ -72,8 +75,6 @@ public class CdiBuilder {
             logger.debug("Could not install our singleton service");
         }
 
-        //TODO there must be a better place to initialize this
-        ThreadContext.addThreadContextListener(new OWBContextThreadListener());
         return SINGLETON_SERVICE;
     }
 
--- a/container/openejb-core/src/main/java/org/apache/openejb/core/ThreadContext.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/core/ThreadContext.java
@@ -24,17 +24,17 @@ import org.apache.openejb.util.Logger;
 
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
-import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.Set;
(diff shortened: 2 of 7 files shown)


container/openejb-core/src/main/java/org/apache/openejb/cdi/CdiBuilder.java
```diff
@@ -37,12 +37,14 @@
...
 
+    private static final RequestScopedThreadContextListener REQUEST_SCOPED_THREAD_CONTEXT_LISTENER = new RequestScopedThreadContextListener();
+
     public void build(final AppInfo appInfo, final AppContext appContext, final List<BeanContext> allDeployments) {
         initSingleton().initialize(new StartupObject(appContext, appInfo, allDeployments));
     }
 
     private ThreadSingletonService initSingleton() {
-        ThreadContext.addThreadContextListener(new RequestScopedThreadContextListener());
+        ThreadContext.addThreadContextListener(REQUEST_SCOPED_THREAD_CONTEXT_LISTENER);
```
brunobat: I'm not comfortable with this change as this serves request scoped beans, hence multiple threads and multiple applications.

container/openejb-core/src/main/java/org/apache/openejb/cdi/CdiBuilder.java
```diff
@@ -62,6 +64,7 @@ public void build(final AppInfo appInfo, final AppContext appContext, final List
 
     public static synchronized ThreadSingletonService initializeOWB() {
         logger.info("Created new singletonService " + SINGLETON_SERVICE);
+        SystemInstance.get().addObserver(SINGLETON_SERVICE);
```
brunobat: Why this?

tomee/tomee-catalina/src/main/java/org/apache/tomee/catalina/TomcatLoader.java
```diff
@@ -187,7 +189,7 @@ public void initialize(final Properties properties) throws Exception {
         System.setProperty("openejb.base", SystemInstance.get().getBase().getDirectory().getAbsolutePath());
 
         // Install tomcat thread context listener
-        ThreadContext.addThreadContextListener(new TomcatThreadContextListener());
+        ThreadContext.addThreadContextListener(threadContextListener);
```
brunobat: Again, not comfortable with static listeners. Is this only instantiated once, globally?

container/openejb-core/src/main/java/org/apache/openejb/core/ThreadContext.java
```diff
@@ -24,17 +24,17 @@
...
-import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class ThreadContext {
 
     private static final Logger log = Logger.getInstance(LogCategory.OPENEJB, "org.apache.openejb.util.resources");
-    private static final ThreadLocal<ThreadContext> threadStorage = new ThreadLocal<ThreadContext>();
-    private static final List<ThreadContextListener> listeners = new CopyOnWriteArrayList<ThreadContextListener>();
```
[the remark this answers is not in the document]
doychin: Order is not relevant. The idea behind use of Set is to have only one instance of the listener in the set if same listener is added multiple times.

That is the main reason why I moved creating of listener instances.

This is only for listeners that does not depend on anything else and just switch the context when thread context is switched.

container/openejb-core/src/main/java/org/apache/openejb/cdi/CdiBuilder.java
doychin: There is no need of multiple instances. RequestScopedThreadContextListener creates the scope and keeps it inside the supplied ThreadContext. 
Also initSingleton is called only once when CDI container is initialized. That container is initialized when first CDI enabled app is deployed. 

When new  instance is created when server is stopped and again started(multiple starts of tests in arquilian) new instance of RequestScopedThreadContextListener  will be added in the original code. Instead now same instance will be "added" to the set of listeners.

This listener is never removed from listener list/set

container/openejb-core/src/main/java/org/apache/openejb/cdi/CdiBuilder.java
doychin: ThreadSingletonService will be notified when it is removed during server shutdown and the code added in ThreadSingletonService will remove the associated ThreadContextListener from listeners set. 
See change in ThreadSingletonServiceImpl

tomee/tomee-catalina/src/main/java/org/apache/tomee/catalina/TomcatLoader.java
doychin: Same reason like above for RequestScopedThreadContextListener 

These ThreadContextListeners don't depend on anything else and they just manipulate the passed data.

container/openejb-core/src/main/java/org/apache/openejb/core/ThreadContext.java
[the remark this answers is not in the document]
brunobat: ok, let's see how it behaves. Thanks

container/openejb-core/src/main/java/org/apache/openejb/cdi/CdiBuilder.java
brunobat: ok

container/openejb-core/src/main/java/org/apache/openejb/cdi/CdiBuilder.java
brunobat: ok

PR #375, opened 2019-01-17
Removed redundant public modifier from ServerService interface - TOMEE-2455

BogdanStirbat:


brunobat:
Will build the container package to see if this doesn't break something.

PR #383, opened 2019-02-02
Add README.adoc to concurrency-utils example: TOMEE-2375

BogdanStirbat:
Jira task: https://issues.apache.org/jira/browse/TOMEE-2375.


BogdanStirbat:
Hi,

Thank you for the feedback! I think that indeed, so much code makes the documentation hard to follow. 

I've reduced the length of the documentation in the following way: I've added link to all the classes, I've reduced the length of listed classes to the bare minimum, I've eliminated all other code fragments, I've cut the 'Running the tests' area.

If you have any other feedback, please tell.

PR #389, opened 2019-02-08
Azure Example - Deployment must happen on port 80. 

brunobat:
The port was commented by mistake in the original PR.
Webprofile is faster to deploy

Signed-off-by: brunobat <brunobat@gmail.com>


cesarhernandezgt:
+1

PR #409, opened 2019-02-22
TomEE-2472 updating mp rest client and cxf

ivanjunckes:


j4fm:
If I remember when I looked a while ago at CXF 3.3 there were some dependency changes. Like activation-1.1.jar gets removed from TomEE - I know some software that depends on that included with TomEE and would break (maybe I'm mistaken).  Does it pass all the tests locally?

PR #418, opened 2019-02-26
Bump geronimo openapi & safeguard to released versions

j4fm:


j4fm:
The released versions are not yet in Maven but soon.  So the PR is waiting but not ready yet.

j4fm:
These are now in maven. So ready to go

PR #427, opened 2019-02-28
 TOMEE-2481 - Move code using PropertyEditors (deprectaed class) to PropertyEditorRegistry

danielsoro:


otaviojava:
@danielsoro every time you need to use the method
` new PropertyEditorRegistry().registerDefaults().`
Why not make this method static?
`PropertyEditorRegistry.registerDefaults();`

danielsoro:
Well.. this class is from xbean-reflect library.

Should exists reason because it is not static and I'm calling it to make sure that we have all default convert registered before get the value. 

That is what PropertyEditor was doing, but now it is totally deprectead just because it is what is recommended now: `this is all static and leaks, use PropertyEditorRegistry`

jgallimore:
In`LocalJMXCommand` the PropertyEditorRegistry is held as a field. In other places, this is doing `PropertyEditorRegistry.registerDefaults()` when calling `getValue()`, which has to go an do a bunch of registration each time.

What do we think the lifecycle of a `PropertyEditorRegistry` should be? I did a quick search for references, which throws up `ManagedMBean`, `ActiveMQ5Factory`, `JMSProducerImpl`, and I would have thought it could be tied to the lifecycle of the components using it in each case (i.e. - make it a field).

Thoughts?

rzo1:
@danielsoro can you resolve the conflicht in `ClaimBean.java` ?

@jgallimore I agree with your thoughs. I also think, that the `PropertyEditorRegistry` should be tied to the lifecycle of the component using it. Should we change this inside this PR or open a new JIRA and do this optimization work in a separate branch / take it as a feature? I would be +1 for the JIRA option.

jeanouii:
Hey @rzo1 
We haven't heard from Daniel.

So if you are willing to help, I suggest you create a new PR, cherry pick what you need from this one and move on.

Does it work for you?

rzo1:
@jeanouii Ok - i will submit a PR soon

jgallimore:
This PR has been merged: https://github.com/apache/tomee/pull/541 - is this one still needed?

rzo1:
@jgallimore Yes. Can be closed.

PR #436, opened 2019-03-07
Ignore temp folder for tck tests

danielsoro:


danielsoro:
This temp folder appear every build.
it can be pushed to the repository accidentally, that because updating the gitignore. :)

PR #451, opened 2019-03-26
Deprecate unused method (tomee-2499)

doychin:
Check https://issues.apache.org/jira/projects/TOMEE/issues/TOMEE-2499

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/assembler/classic/Assembler.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/assembler/classic/Assembler.java
@@ -2259,22 +2259,6 @@ public class Assembler extends AssemblerTool implements org.apache.openejb.spi.A
         }
     }
 
-    public void destroyApplication(final AppContext appContext) throws UndeployException {
-
-        final ReentrantLock l = lock;
-        l.lock();
-
-        try {
-            final AppInfo appInfo = deployedApplications.remove(appContext.getId());
-            if (appInfo == null) {
-                throw new IllegalStateException(String.format("Cannot find AppInfo for app: %s", appContext.getId()));
-            }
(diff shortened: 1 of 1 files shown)


container/openejb-core/src/main/java/org/apache/openejb/assembler/classic/Assembler.java
```diff
@@ -2259,22 +2259,6 @@ public void destroyApplication(final String filePath) throws UndeployException,
         }
     }
 
-    public void destroyApplication(final AppContext appContext) throws UndeployException {
```
AndyGee: I'd never remove a public method like this, as it's impossible to tell if some other tool is using this.

AndyGee: I'd actually just write the 'missing' test

doychin: This method will never work. It will always fail with IllegalStateException. Deployed applications are kept in a map and the key is the path of the deployed application. Path information is not available in AppContext and in this method application is searched by appContext.getId() which is not app path. 

That is the main reason I suggest this method to be removed.

AndyGee: I agree it's useless. But I'd still just deprecate it for the next release, then remove it after. I just can't guarantee that some 3rd party has not written code around it, even if it throws an error.

rzo1:
@dblevins This PR "only" adds a `@Deprecated` - I think, this is fine (no impact on the existing code-base), see https://issues.apache.org/jira/projects/TOMEE/issues/TOMEE-2499

@doychin Shouldn't we backport it for 7.0.x / 7.1.x ?

PR #452, opened 2019-03-26
Make ArrayEnumeration generics aware

doychin:
Also cleanup code in HttpRequestImpl.


AndyGee:
@jgallimore wdyt?

rzo1:
We have a similar class `ArrayEnumeration` in `org.apache.openejb.client`. Shouldn't we change it there too?

Besides that, I agree with @AndyGee : Some syntax changes -> shouldn't be a problem.

PR #457, opened 2019-04-10
TOMEE-2507

tandraschko:


jgallimore:
Hi! Thanks for the PR! I don't have any specific issues, but out of curiosity, what's the driver for these excludes?

jfreechart-*,jscience-*,lesscss-engine-*,rhino-*

tandraschko:
It's an sub-task for my peformance improvements :)
TomEE 7.x still takes the half time to startup on my applications as 8.0.0-SNAPSHOT and i try ti improve it further.

This are known libraries but probably not so often used in TomEE applications.

tandraschko:
I also wonder if we could skip (cc @rmannibucau):
jackson-datatype-jsr310-
jackson-jaxrs-json-provider-
jackson-jaxrs-base-

rmannibucau:
LGTM

tandraschko:
@jgallimore should i provide another PR for 
jackson-datatype-jsr310-
jackson-jaxrs-json-provider-
jackson-jaxrs-base-
?

PR #459, opened 2019-04-14
[TOMEE-2509] Wrap FileWriter with a BufferedWriter in a loop

bd2019us:
Fixes ISSUE [#TOMEE-2509](https://issues.apache.org/jira/browse/TOMEE-2509)

BufferedWriter can effectively reduce the times of IO access with caches, especially in a loop, which will be the performance bottleneck of program.


cesarhernandezgt:
Hi @bd2019us 

Thank you for your contribution, in order to move forward this PR please open an email thread on the mailing list to make reference to both the Jira and the PR. 

In this way, we collaborate with the committers to move forward the contributions and follow the Apache Way "If it didn't happen on a mailing list, it didn't happen." [1] along with the recommended Apache TomEE Workflow [2].

[1] https://community.apache.org/newbiefaq.html#NewbieFAQ-IsthereaCodeofConductforApacheprojects?
[2] http://tomee.apache.org/community/contributing/workflow.html

bd2019us:
Hi @cesarhernandezgt , thanks for your kind reply and I will work on it as soon as possible.

PR #465, opened 2019-05-07
TOMEE-2515 MicroProfile JWT 1.1 Cleanup

dblevins:
Still in the trimming code stage.  Plan to get on to logging improvements and greater tests shortly.


mp-jwt/src/main/java/org/apache/tomee/microprofile/jwt/config/JWTAuthConfiguration.java
```diff
@@ -29,47 +29,47 @@
...
  */
-public class JWTAuthContextInfo {
-    private static final Logger logger = Logger.getLogger(JWTAuthContextInfo.class.getName());
+public class JWTAuthConfiguration {
+    private static final Logger logger = Logger.getLogger(JWTAuthConfiguration.class.getName());
     public static final String DEFAULT_KEY = "DEFAULT";
 
-    private Map<String, Key> signerKeys;
-    private String issuedBy;
+    private Map<String, Key> publicKeys;
```
jeanouii: Not so sure about this one.
Key is the base class and can be a public key or a secret key.

What about signingKey?
It would be valid for either symmetric or asymmetric key.

dblevins: Changed to "verificationkey" most places

dblevins:
Merging so we can get a build from the CI and move forward with the 8.0.0-M3 release

PR #454, opened 2019-04-01
Fixes  Eclipse Microprofile-Open-API when it reads Application. [TOMEE-2502]

otaviojava:
This PR fixes the Eclipse Microprofile Open-API.
Briefly, at the current version, TomEE does not read information from the `Application` class that has Open-API annotation.

```
@ApplicationPath("api")
@OpenAPIDefinition(info = @Info(
        title = "Example application",
        version = "1.0.0",
        contact = @Contact(
                name = "Otavio",
                email = "otavio@otavio.com",
                url = "http://www.otaviojava.com.br")
),
        servers = {
                @Server(url = "/example", description = "localhost")
        }
)
public class MVCApplication extends Application {
}
```

https://issues.apache.org/jira/browse/TOMEE-2502

PR #467, opened 2019-05-16
TOMEE-2502 + Test

jgallimore:
Building on https://github.com/apache/tomee/pull/454 - this adds a test and looks at the impact of removing `SystemInstance.get().setProperty("openejb.cxf-rs.cache-application", "false");`


jgallimore:
This currently has a handful of test failures, so may need some modification.

jgallimore:
I think we can look at the system property in another PR. With the test, this looks like its ok to go. Just running a check and then I'll merge this in.

jgallimore:
Even after putting back `SystemInstance.get().setProperty("openejb.cxf-rs.cache-application", "false");`, this still has multiple failures across health, metrics and OpenAPI. I'll retest it again, but I'd be grateful for another pair of eyes.

PR #479, opened 2019-06-02
Create java-compatibility.adoc

dineshkumarg:
Maintaining compatibility helps users to know Java versions support for TomEE

### DIFF
--- a/docs/java-compatibility.adoc
+++ b/docs/java-compatibility.adoc
@@ -1,0 +1,4 @@
+= Java Compaibility with TomEE 7.1.x
+:index-group: General Information
+
+TomEE 7.1.0 is Java EE 7, Java 8 and limited MicroProfile 1.2 (JWT, Config, HealthCheck).  Any OS that supports Java 8 works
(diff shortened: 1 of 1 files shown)


cesarhernandezgt:
I think we can iterate after this PR to actually provide compatibility Matrix into TomEE branches :)
The information is right in the poms so for example, we can provide MicroProfile specific version, Cfx version etc:

https://github.com/apache/tomee/blob/8c6358cca46e431df78fc9c7d818259a9ba8635b/pom.xml#L198-L210

The overall goal is to provide a user friendly version matrix without the need to go into the repository and pom.xml files. We can even automate the process but that could be a third iteration.

jeanouii:
Thank you so much guys for the doc and the collaboration on the document.
Going to merge it

jeanouii:
How about creating another PR for TomEE 8.x (aka master)?
We could update it on a regular basis

dineshkumarg:
> How about creating another PR for TomEE 8.x (aka master)?
> We could update it on a regular basis

Sure, I will create a similar document for TomEE 8.x

Thank you

rmonson:
Hi,
Dinesh.  Do you have any more contact information?  Like an email address?

dineshkumarg:
> Hi,
> Dinesh. Do you have any more contact information? Like an email address?

yes it's dinesh.gaddam@gmail.com

PR #481, opened 2019-06-08
Extract some of the values in the test into constant definitions.

doychin:
Reduce the size of the pool to very small amount so that when test is run on high core CPU's it will not fail.


jeanouii:
I am happy to merge this one. I'd like to learn more about the failures though.

In real world, TomEE is meant to be used on high core CPU's even though the user might configure the pool with a bigger value than 5.

If it starts failing it's not good.
So I want to make sure we are not hiding a bug or a concurrency issue.

Thanks

doychin:
The failure happens when the CPU used to run the test has many cores and is fast - like many new 8 or more core desktop CPU's 

In this case the code that receives the messages executes fast enough so a load of 100 messages is not enough to fill pool size of 30 message receivers. It reuses the first 20-25 instances. The end value in the test varies. 

The problem becomes visible when you run the same test multiple times after JVM has enough time to optimize the test code.

That is the main reason I set such low value for max pool size.

PR #483, opened 2019-06-10
Add cleanup before generating keys. Make sure new keys are generated …

doychin:
…for every test run.


jeanouii:
Thanks guys, it looks good.

PR #487, opened 2019-06-14
Added Grinder setup for XA performance test.

cesarhernandezgt:


jgallimore:
I expanded on this in https://github.com/apache/tomee/pull/488

PR #488, opened 2019-06-18
Performance tests for XA example

jgallimore:
Follow on from #487. Strips out all jars from the project/repository, and creates a Grinder directory when running the Maven build. Uses Apache Derby out of the box. Some tweaks to the datasource config for the tests and runtime resource definitions.


examples/xa-datasource/src/main/webapp/WEB-INF/resources.xml
```diff
@@ -0,0 +1,46 @@
+<resources>
+
+  <Resource id="movieDatabase" type="DataSource">
```
cesarhernandezgt: This is how I manage to configure XA datasource on Derby embedded.

```
  <Resource id="movieDatabase" type="DataSource">
    XaDataSource  movideXAResource
    JdbcDriver  org.apache.derby.jdbc.EmbeddedDriver
    JdbcUrl  jdbc:derby:testdb;create=true
    UserName admin
    Password admin
    ValidationQuery=values 1
    InitialSize=2
    MaxActive=128
    MaxIdle=25
    MinIdle=10
    TestWhileIdle=true
    TestOnBorrow=false
    TestOnReturn=false
    AccessToUnderlyingConnectionAllowed=true
    TimeBetweenEvictionRuns=1 minute
    MaxWaitTime=0 seconds
    JtaManaged=true
  </Resource>

  <Resource id="movideXAResource" type="XADataSource" class-name="org.apache.derby.jdbc.EmbeddedXADataSource">
    user admin
    Password admin
    DatabaseName testdb
    CreateDatabase create
  </Resource>
```

With the above configuration for `movieDatabase` data source, I only get the following warning:
```
 WARNING [main] org.apache.openejb.assembler.classic.Assembler.unusedProperty unused property 'InitializeAfterDeployment' for resource 'dbtest/movideXAResource'
```

But I think this is added on the fly by TomEE since for `movideXAResource` we specify that the database should be created `CreateDatabase create` and when I removed this property, the `movideXAResource` complained that there were no database available.

PR #506, opened 2019-07-08
TOMEE-2564 Classpath with :: on Linux is cause an infinite loop on War deploy

cbos:
See ticket https://issues.apache.org/jira/browse/TOMEE-2564 for all details of the issue


jgallimore:
Good catch. Is there a way we could test the change without introducing an infinite loop? I'm also wondering if we could/should check the file exists, and perhaps add log some warnings. What do you think?

jgallimore:
Merged this in, but the feedback I mention in my previous comment would be a welcome additional contribution. Thanks for the PR!

cbos:
> Good catch. Is there a way we could test the change without introducing an infinite loop? I'm also wondering if we could/should check the file exists, and perhaps add log some warnings. What do you think?

An exists() check on the file before converting the file to URL could have solved the issue as well. It returns false. 

Writing a test might be possible. At this moment there is in the class TomEEJarScannerTest only a test for the TomEEFilter, I did not try to have a test for TomEEJarScanner.
It requires tweaking of the java.class.path System property in the test (Add :: via setProperty and revert to the old value in the @after).

PR #507, opened 2019-07-09
TOMEE-2534 - Translate to Portuguese: examples/access-timeout.

gabrielferreirapro:
Hey guys, there is my translation to examples/access-timeout.
I had enough difficulty with this example because some terms like "vendor", "bean", "container" and others, I don't know if it's lack of experience on my part but I see more this terms in english and its very difficult to translate literally. 
I did not want to dare to translate the specification into the part in which it is quoted so as not to run the risk of misinterpreting it.
I am totally open to suggestions for improvement.
Thank you !
Jira task: https://issues.apache.org/jira/browse/TOMEE-2534

### DIFF
--- a/examples/access-timeout/README_pt.adoc
+++ b/examples/access-timeout/README_pt.adoc
@@ -1,0 +1,217 @@
+= Anotação @AccessTimeout
+:index-group: EJB
+:jbake-type: page
+:jbake-status: published
+
+Antes de darmos uma olhada no `@AccessTimeout`, pode ajudar entender quando uma chamada precisa "esperar".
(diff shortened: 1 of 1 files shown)


Daniel-Dos:
About the terms, you do not need to translate them literally, as it can bring confusion to anyone who reads. 

great translate. thanks :  )

gabrielferreirapro:
> About the terms, you do not need to translate them literally, as it can bring confusion to anyone who reads.
> 
> great translate. thanks : )

Make sense Daniel, thank you for the tip!

PR #532, opened 2019-08-06
TOMEE-2612 - updated  javaee-api to 8.0-1

cesarhernandezgt:


rzo1:
We need to resolve MR conflics, I guess :)

rzo1:
We are on the way to `8.0-2` so this PR needs an update :)

rzo1:
As `8.0-2` is ready now, we should update this PR and the JIRA to reflect latest changes :)

jgallimore:
@cesarhernandezgt Is this PR still needed?

cesarhernandezgt:
> @cesarhernandezgt Is this PR still needed?

We can close this PR. But I'll update TomEE since it seems now apart from the version we also need to fix the maven version format: https://github.com/apache/tomee/blob/master/examples/access-timeout-meta/pom.xml#L57

PR #540, opened 2019-08-21
TOMEE-2631 - Update Deltaspike version to 1.9.1

Daniel-Dos:
link to jira  = https://issues.apache.org/jira/browse/TOMEE-2631


Daniel-Dos:
I need see that.

I must has made a copy for accident.

Thanks for review.

PR #547, opened 2019-09-03
Add jgallimore's JMS Tests and fix TomEE-2506 autoconnection tracker warning

exabrial:


jgallimore:
There's a few conflict on here, but it looks good. I'll get it merged.

PR #550, opened 2019-09-06
TOMEE 2657 - translation into pt-br the examples/vaadin-lts-v08-simple

marcoantoniobferreira:


examples/vaadin-lts-v08-simple/README_pt.adoc
```diff
@@ -0,0 +1,56 @@
+== Vaadin V8 (LTS) - WebApp Simples em Java
```
cesarhernandezgt: Hi @marcoantoniobferreira , thanks for the PR.
I'll suggest adding the document headers to define jbake metadata and index-group.
Check for instance: https://github.com/apache/tomee/blame/master/examples/mp-metrics-gauge/README.adoc#L3-L5

marcoantoniobferreira: Hi @cesarhernandezgt thanks for review. In this case, which is the correct index-group?

cesarhernandezgt: Currently, appears on Uknown section:
image

IMHO, I think it should be under Frameworks:
image

If you want to test the end result of the index-group, you will need to run the website locally on your machine to make sure it appear correctly: https://github.com/apache/tomee-site-generator/

PR #556, opened 2019-09-07
[TOMEE-2669] [DOC] [Translation] Translate async-methods example to Spanish

cchacin:
https://issues.apache.org/jira/browse/TOMEE-2669


dblevins:
@cchacin Thank you for the careful update of the English and Portuguese versions too! :)

PR #560, opened 2019-09-07
TOMEE-2383

sercheo87:
Add documentation of example groovy jpa.


dblevins:
@sercheo87 You're on fire today :)

sercheo87:
> @sercheo87 You're on fire today :)

To kill time :smile::smile:

PR #576, opened 2019-09-29
TOMEE-2696 Complete mp-config-source-database Example's Explanation

emecas:

PR #577, opened 2019-09-29
TOMEE-2695 Add Spanish translation for mp-config-source-database

emecas:
Add Spanish translation for:

https://github.com/apache/tomee/blob/master/examples/mp-config-source-database/README.adoc

Since the original readme.adoc was kind of incomplete, I've extended it in  #576  (TOMEE-2696)

### DIFF
--- a/examples/mp-config-source-database/README_es.adoc
+++ b/examples/mp-config-source-database/README_es.adoc
@@ -1,0 +1,109 @@
+= MicroProfile Configuration ConfigSource Database
+:index-group: MicroProfile
+:jbake-type: page
+:jbake-status: published
+
+Este es un ejemplo sobre cómo implementar un ConfigSource para la configuración personalizadade microprofile. El ConfigSource personalizado lee los valores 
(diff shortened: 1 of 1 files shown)


examples/mp-config-source-database/README_es.adoc
```diff
@@ -0,0 +1,109 @@
+= MicroProfile Configuration ConfigSource Database
+:index-group: MicroProfile
+:jbake-type: page
+:jbake-status: published
+
+Este es un ejemplo sobre cómo implementar un ConfigSource para la configuración personalizadade microprofile. El ConfigSource personalizado lee los valores
```
breakponchito: I think here you forget the space between 'personalizada' and 'de' words

emecas: Good catch, thanks, it is fixed now!

MauroVelasco: Also update MicroProfile instead of microprofile could be cool

PR #588, opened 2019-10-15
TOMEE-2711 translation into portuguese examples/vaadin-lts-v10 and fix in jbake english version

marcoantoniobferreira:


monsonhaefel:
I looked it over in Google translate and it looks good to me!

PR #590, opened 2019-10-17
[TOMEE-2494] Fix Translation to MicroProfile in mp-config-source-database example

emecas:


monsonhaefel:
According to Google translate this change looks good. Let's merge it!

PR #599, opened 2019-10-26
[TOMEE-2723] Fix myfaces-codi-demo Example and Add Spanish Translation

emecas:
https://issues.apache.org/jira/browse/TOMEE-2723


rmonson:
Looks good to me as well. Let's merge it!

PR #604, opened 2019-11-11
TOMEE-2730 add mp.jwt.tomee.allow.no-exp option

jgallimore:
This change adds the mp.jwt.tomee.allow.no-exp option (it is false by default). If set to true, the exp claim will be validated when it is present, but the JWT token will not fail validation if the claim is missing - i.e. the token will be valid forever.


rmonson:
It looks pretty straight forward. I did notice that the change to JWTAuthContextInfoProvider. getOptionalContextInfo() always returns false for the AllowNoExpiryClaim.  Just want to check that that was intended. No explanation required if it was intended.

jgallimore:
That is intended. That particular change is in test code as opposed to production code.

PR #608, opened 2019-11-13
Ddded missing dependency for microprofile

scriptmonkey:
Jira card:  TOMEE-2731

This fixes a missing dependency for microprofile.  This was discussed in the dev email list.

I added:
```
<dependency> <!-- Added in for backwards compatibility with the move to Jakarta EE coordinates -->
        <groupId>com.sun.activation</groupId>                                      
        <artifactId>jakarta.activation</artifactId>                                
        <version>1.2.1</version>                                                   
</dependency> 
```

### DIFF
--- a/tomee/tomee-microprofile/mp-common/pom.xml
+++ b/tomee/tomee-microprofile/mp-common/pom.xml
@@ -158,6 +158,12 @@
       </exclusions>
     </dependency>
 
+    <dependency> <!-- Added in for backwards compatibility with the move to Jakarta EE coordinates -->
+      <groupId>com.sun.activation</groupId>
+      <artifactId>jakarta.activation</artifactId>
+      <version>1.2.1</version>
+    </dependency>
+
     <dependency>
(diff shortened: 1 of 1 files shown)


tomee/tomee-microprofile/mp-common/pom.xml
```diff
@@ -158,6 +158,12 @@
       </exclusions>
     </dependency>
 
+    <dependency> <!-- Added in for backwards compatibility with the move to Jakarta EE coordinates -->
```
rmannibucau: looks like you just want to copy the same than in openejb-core for consistency + use provided scope since on j8 it must not be transitive

see https://github.com/apache/tomee/blob/master/container/openejb-core/pom.xml#L558

scriptmonkey:
I will fix and make a new pull request.

jgallimore:
You could just make the change in your branch, commit and push again (which will update the PR here). Alternatively we could merge this, and you follow up with the refactor - that's also fine. New PR is also good. Shout if you get stuck, and thanks for giving this a go.

PR #613, opened 2019-11-22
TOMEE-2744: JD14 compatibility: remove usage of javax.security.acl

cocorossello:
Some tests in openejb-core fail, but it doen't seem related at all


cocorossello:
Hi,

In the code acl.Group is used basically as a Principal with members support, is not used as a Policy, as far as I can see. 

With this change our environment works  in jdk11, 12, 13 and 14, I don't see any reason why it wouldn't work in jdk8 (but of course I don't know all the use cases here).

jgallimore:
I'll merge this in, and we'll let the CI build it. Hopefully all will look good, but we'll revisit it if we see tests failing.

PR #619, opened 2019-12-07
[TOMEE-2749] Add Spanish translation for mp-faulttolerance-retry

emecas:


jgallimore:
@cesarhernandezgt can you review when you get time?

PR #623, opened 2019-12-17
Tomee 2756

kubo44:
Small fix of the typo in the DeploymentsResolver class which caused excluding all URLs.


kubo44:
Sorry I used wrong from branch (master), therefore so many changes..

PR #626, opened 2020-01-03
TOMEE-2759-Update-Eclipse-Krazo

Daniel-Dos:
Link to Jira = https://issues.apache.org/jira/browse/TOMEE-2759

### DIFF
--- a/examples/mvc-cxf-hibernate/pom.xml
+++ b/examples/mvc-cxf-hibernate/pom.xml
@@ -31,8 +31,8 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <tomee.version>8.0.2-SNAPSHOT</tomee.version>
-    <version.deltaspike>1.9.1</version.deltaspike>
-    <version.krazo>1.0.0-RC1</version.krazo>
+    <version.deltaspike>1.9.3</version.deltaspike>
+    <version.krazo>1.1.0-M1</version.krazo>
     <version.arquillian>1.1.13.Final</version.arquillian>
     <version.graphene.webdriver>2.3.1</version.graphene.webdriver>
   </properties>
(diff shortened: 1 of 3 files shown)


Daniel-Dos:
thanks @cesarhernandezgt

>

cesarhernandezgt:
Hi @Daniel-Dos 
I couldn't merge the PR since it seems there is a small conflict.
I'm happy to review again after that gets fixed.

emecas:
This could be a little off-topic [or not :)] 

Since this upgrade only applied for those 3 examples, is there a way to include a check to validate the tests for each example are still passing , or is something like that already on place? or even covered and implicit to the general build?

Should not be the process for upgrading dependencies a more rigorous process ? at least from the point of view of documenting,  or letting know what is the specific benefit of doing it, or why it is important, or by the simple fact of checking and ensure full backward compatibility.

Also it could be part of the description of the Dependency upgrade JIRA issue?

By the way there is already a new release after the one included in this PR.
https://github.com/eclipse-ee4j/krazo/releases/tag/1.1.0-M1

cesarhernandezgt:
> Since this upgrade only applied for those 3 examples, is there a way to include a check to validate the tests for each example are still passing, or is something like that already on place? or even covered and implicit to the general build?

Currently, if at least one example fails on either compilation, test or build, the entire TomEE build fails since `examples` is a maven submodule.

> Should not be the process for upgrading dependencies a more rigorous process ? at least from the point of view of documenting, or letting know what is the specific benefit of doing it, or why it is important, or by the simple fact of checking and ensure full backward compatibility.
> 
> Also it could be part of the description of the Dependency upgrade JIRA issue?

Good observation. Usually, I have seen the other approach: An issue who's bug gets resolved by a dependency update. But I agree that in both cases the context of the dependency upgrade should be reflected on the Jira.

Daniel-Dos:
@cesarhernandezgt ,

I'm going to close this PR and open another one, because there was a problem in the conflict merge.

PR #628, opened 2020-01-30
TOMEE-2767 / Deleting typos and correcting heading syntax

bitcod3r:
I confirm that:
- Code compiles with JDK 8.
- adoc file produces a correct output with asciidoc.
- And, all tests pass successfully.

This is my first PR to the TomEE project. Please, give me your feedback to know what might I improve.  

More PRs are coming... :)

### DIFF
--- a/docs/documentation.adoc
+++ b/docs/documentation.adoc
@@ -9,31 +9,29 @@ See also the link:examples-trunk/index.html[examples page] for
 downloadable, executable and code-focused view of Java EE and TomEE. You
 can also find us on IRC freenode.org #openejb and #tomee
 
-\{row
-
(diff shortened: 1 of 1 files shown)


djencks:
I think this is a fantastic change, thanks!

Personally, I think the page would look even better with the links in unordered lists rather than paragraphs with line breaks. Does anyone agree?

bitcod3r:
> Personally, I think the page would look even better with the links in unordered lists rather than paragraphs with line breaks. Does anyone agree?

+1 @djencks,  I agree with you. It is semantically correct to markup those links as (unordered) lists of resources. Adding 1 commit to the current PR.

PR #630, opened 2020-03-08
TOMEE-2786 Spanish translation for vaadin-vxx-simple example

emecas:
https://issues.apache.org/jira/browse/TOMEE-2786

I noted that this example documentation emphasizes about using the latest vaadin release available, that currently are  alpha: `14.2.0.alpha6`  and stable versions: `14.1.18`  and  `15.0.0`

But the `pom.xml` file is using `14.0.0.rc1` updated about ~9 months ago by  commit 9fd13f96809

Would we update vaadin version for the examples as well at this time?

Ref: https://github.com/vaadin/platform/releases

### DIFF
--- a/examples/vaadin-vxx-simple/README_es.adoc
+++ b/examples/vaadin-vxx-simple/README_es.adoc
@@ -1,0 +1,51 @@
+:index-group: Frameworks
+:jbake-type: page
+:jbake-status: published
+
+== Actual Plataforma Vaadin  - Simple Aplicación Web en Java
+
+Este ejemplo muestra como iniciar con una simple aplicación web que usa Vaadin Flow,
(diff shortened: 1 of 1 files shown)


emecas:
@cesarhernandezgt may you take a look of this PR?

cesarhernandezgt:
@emecas 
Hi, I agree and I vote +1 on update from vaadin `14.0.0.rc1` to `14.1.18`.
Also if there is no braking change, `15.0.0` sounds like a more appealing version to use :).

emecas:
@cesarhernandezgt @Daniel-Dos  Thank you. I will add soon a new full example using the most recent&compatible Vaadin version available.

Daniel-Dos:
very good :  )

PR #662, opened 2020-06-12
Employ the TomEE Patch Plugin

dblevins:
Ok, this is working and looks to be in good shape.  Server boots with no issues and can display the splash page.  Haven't yet been able to test other aspects.


jgallimore:
LGTM!

PR #673, opened 2020-07-09
fix EndpointHandlerTest.java test

cesarhernandezgt:


jgallimore:
Good catch, thank you!

PR #692, opened 2020-07-21
TOMEE-2809: Update to OWB 2.01.6

cocorossello:
Update straight to 2.0.17 should be better. It's done in https://github.com/apache/tomee/pull/693

I'll close this PR

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1460,6 +1460,11 @@
         </exclusions>
       </dependency>
       <dependency>
+        <groupId>org.apache.xbean</groupId>
+        <artifactId>xbean-asm-util</artifactId>
+        <version>${xbeanVersion}</version>
+      </dependency>
+      <dependency>
         <groupId>org.codehaus.swizzle</groupId>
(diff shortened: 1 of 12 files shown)


pom.xml
```diff
@@ -115,15 +115,15 @@
     <geronimo-javamail_1.4_mail.version>1.9.0-alpha-2</geronimo-javamail_1.4_mail.version>
 
     <openjpa.version>3.1.0</openjpa.version>
-    <openwebbeans.version>2.0.12</openwebbeans.version>
+    <openwebbeans.version>2.0.15</openwebbeans.version>
```
rmannibucau, line 118: ticket mentions 2.0.16 and last release is 2.0.17 so maybe directly go to 2.0.17?

cocorossello: I thought it would be easier to do it in two steps (at least for me), since 2.0.17 comes with the asm8 change. I was planning to do it once this one is finished

cocorossello: Sorry, this should be 2.0.16, I will push this now

PR #699, opened 2020-08-01
TOMEE-2892 Add Spanish Translation: injection-of-entitymanager

emecas:


Daniel-Dos:
thanks @emecas for translate : )

PR #716, opened 2020-11-06
TOMEE-2917 - Change All WrappingMessage to become serializable

tingung:
JIRA: https://issues.apache.org/jira/browse/TOMEE-2917

Fix Summary: Change all WrappingMessage to implements Serializable and add unique serialVersionUID. 

Notes:
There is a business use case which need to persist WrappingTextMessage (or TextMessage) into database. Required WrappingTextMessage to become serializable.
TextMessage implementation by both IBM Websphere and Weblogic are serializable.
https://www.ibm.com/support/knowledgecenter/en/SSFKSJ_7.5.0/com.ibm.mq.javadoc.doc/WMQJMSClasses/com/ibm/jms/JMSMessage.html
https://docs.oracle.com/middleware/12213/wls/WLAPI/weblogic/jms/common/TextMessageImpl.html


rzo1 (approved):
If `java.io.Serializable` is implemented here, you should add a corresponding `serialVersionUID`.

tingung:
> If `java.io.Serializable` is implemented here, you should add a corresponding `serialVersionUID`.

ok. let me add serialVersionUID.

rmannibucau:
side note: this is true for all messages of this package so can be worth doing it at once to avoid 5 issues + needs an issue on https://issues.apache.org/jira/projects/TOMEE to be in the changelog ;)

tingung:
> side note: this is true for all messages of this package so can be worth doing it at once to avoid 5 issues + needs an issue on https://issues.apache.org/jira/projects/TOMEE to be in the changelog ;)

Ok. Then I will change all to implement Serializable. Thanks for prompt response.

rzo1 (approved):

rzo1:
@tingung Do you want to create an issue yourself via https://issues.apache.org/jira/projects/TOMEE  or shall I do this for you?

tingung:
@rzo1 , This is my first issue. Let me try to create JIRA ticket. I am not aware of the process.

rzo1:
> @rzo1 , This is my first issue. Let me try to create JIRA ticket. I am not aware of the process.

Feel free :)

tingung:
@rzo1 , I have created JIRA ticket and linked it to this issue.

Daniel-Dos (approved):

rmannibucau:
Maybe a trivial test case ensuring it is actually serializable is worth it, something around https://github.com/apache/tomee/blob/master/container/openejb-core/src/test/java/org/apache/openejb/util/proxy/LocalBeanProxySerializationTest.java in https://github.com/apache/tomee/blob/master/container/openejb-core/src/test/java/org/apache/openejb/activemq/JMS2AMQTest.java (to reuse jms setup) looks not that hard and will prevent regressions at very least.

side note to tomee committers: got in touch with some amq guys and amq (5) should get JMS 2 support so all this code can be dropped at some point (tomee can also push it to amq to make it happen faster).

rzo1:
I have added a simple unit test  (see https://github.com/rzo1/tomee/commit/b3fdbc920955496ba42468abd39c152e6e27662c) based on your branch.

It seems, that the messages from amq do **not** implement serializable. Subsequently, trying to serialize a wrapping message will fail.

rzo1:
The related PR on the AMQ site was closed https://github.com/apache/activemq/pull/570, so AMQ messages will not become **serializable** in the near future:

> This is not necessary and is not going to work as there are many subtypes in those messages that are not serializable. You can use OpenWireFormat to marshall/unmarshall the data into bytes and persist it to a database. See https://github.com/apache/activemq/blob/master/activemq-client/src/main/java/org/apache/activemq/openwire/OpenWireFormat.java
> 
> Also, in the future a Jira needs to be created and tests need to be added.

Thus, I would propose to close this PR as well as the related JIRA as "wont fix".

rmannibucau:
Side note if it helps: all messages can become serializable implementing writeReplace callback since JMS messages are of serializable types (String, byte[], Map, serializable mainly) so it is mainly about having a MessageReplacer returned by writeReplace and a writeReplace in the MessageReplacer to get back an AMQ instance (or not since we just need to respect the JMS API so personally I would just make one MessageReplacer per JMS message type and that's it).

PR #720, opened 2020-11-27
TOMEE-2930 - MoviesArquillianHtmlUnitTest -> Invalid byte 2 of 2-byte UTF-8 sequence. error.

rzo1:
# What does this PR do?

As discussed on the mailing list [3], we found test errors in `MoviesArquillianHtmlUnitTest`. 

While trying to reproduce it on my Linux machine, I found, that the dependencies used for setting up the unit tests do not match with the current project version. 

CI does not cover this as it can retrieve the old snapshot artifacts (either from local disk cache or from an apache snapshot repo).

With the version adjustments, the tests can be executed and do not throw the exception mentioned above

Question, @cesarhernandezgt : Can we run this PR on the CI environment to check, if the issue persists?

# References

- [1] https://issues.apache.org/jira/browse/TOMEE-2930
- [2] https://ci-builds.apache.org/job/Tomee/job/tomee-7.0.x/org.superbiz$moviefun-functional-test/3/testReport/junit/org.superbiz.moviefun/MoviesArquillianHtmlUnitTest/org_superbiz_moviefun_MoviesArquillianHtmlUnitTest/
- [3] https://openejb.markmail.org/search/?q=type:development#query:type%3Adevelopment+page:1+mid:yagkenymmvojwqvh+state:results

### DIFF
--- a/examples/simple-ear/moviefun-functional-tests/pom.xml
+++ b/examples/simple-ear/moviefun-functional-tests/pom.xml
@@ -34,6 +34,19 @@
     <version.shrinkwrap.resolver>2.0.0</version.shrinkwrap.resolver>
   </properties>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <systemPropertyVariables>
+            <project.version>${project.parent.version}</project.version>
+          </systemPropertyVariables>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
   <dependencyManagement>
     <dependencies>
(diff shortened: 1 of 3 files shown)


rmannibucau:
@rzo1 what about adding in surefire plugin the system property "project.version" and just do a System.getProperty("project.version", "default value SNAPSHOT") in the code? will stay inlined this way

rzo1:
> @rzo1 what about adding in surefire plugin the system property "project.version" and just do a System.getProperty("project.version", "default value SNAPSHOT") in the code? will stay inlined this way

Good idea - will resolve such issues in the future. Will update.

rmannibucau (approved):

Daniel-Dos (approved):

PR #725, opened 2020-12-04
TOMEE-2934 Handle public inner classes when looking for a class in classpath

giangianoulas:
When trying to start up the ejbContainer `EJBContainer.createEJBContainer`
If a bean contains an inner class, used like this:
```java
@Asynchronous
	@TransactionAttribute(TransactionAttributeType.REQUIRES_NEW)
	public Future<Contract> processAsynchronously(Contract r) 
```
Then you get an Exception stating that:
```Cannot locate the class {0} from the codebase [{1}]```
This attept to handle this case, if it also fails then the original exception is thrown.

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/config/rules/ValidationBase.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/config/rules/ValidationBase.java
@@ -149,8 +149,17 @@ public abstract class ValidationBase implements ValidationRule {
         final ClassLoader cl = module.getClassLoader();
         try {
             return Classes.forName(clazz, cl);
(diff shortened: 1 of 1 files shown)


rzo1:
Thanks for this PR. Could you fill a JIRA for this: https://issues.apache.org/jira/projects/TOMEE/issues?

Daniel-Dos (approved):

rmannibucau:
Did you identify why the classname didnt have a dollar but a dot? Is it a configuration or scanning issue? It would be better to fix this rather than the consequence (since other code can fall into that issue elsewhere then).

giangianoulas:
@rmannibucau Of course you are correct and this is definetely not the way to go.
I will close this PR

PR #730, opened 2020-12-10
Experimental change to shade taglibs, xalan and serializer

jgallimore:
This should allow the Taglibs library to use the XPath API part of Xalan that it is tightly coupled to, but allow applications to use JAXP provided by the JDK.


jeanouii (approved):
Looks good. Let's see how the TCK will look like after

PR #744, opened 2021-01-08
fix RAT checks when building locally with .idea

cesarhernandezgt:


cesarhernandezgt:
closing since it contains a leaked commit from another branch :/

PR #746, opened 2021-01-09
updated boms pom.xml files via org.apache.tomee.bootstrap.GenerateBoms to remove snapshot dependencies

cesarhernandezgt:

PR #747, opened 2021-01-09
fix RAT checks when building locally with .idea

cesarhernandezgt:


cesarhernandezgt:
The Check in Github actions is failing because `org.apache.tomee.bom:tomee-plus:jar:8.0.6-SNAPSHOT: The following artifacts could not be resolved: org.apache.johnzon:johnzon-core:jar:1.2.9-SNAPSHOT`.

This error was fixed in a separate PR: https://github.com/apache/tomee/pull/746

rzo1 (approved):

Daniel-Dos (approved):

jeanouii (approved):

PR #751, opened 2021-01-18
TOMEE-2927 Translate to Portuguese: examples/serverless-tomee-mp

Daniel-Dos:
Translate to Portuguese the example serverless-tomee-microprofile

link to Jira: https://issues.apache.org/jira/browse/TOMEE-2927


cesarhernandezgt (approved):
+1 from structure perspective since I don't speak Portuguese :)

PR #752, opened 2021-01-18
TOMEE-2960 added basic github actions for tomee-7.1.x branch

cesarhernandezgt:


Humbedooh:
testing jira...ignore me :)

dfoulks1:
please ignore, testing jira fix

dfoulks1:
Jira worklog test comment, please disregard.

rzo1 (commented):
On the `master`, we define a separate step

```
    steps:
    - uses: actions/checkout@v2
    - name: Set up JDK 1.8
      uses: actions/setup-java@v1
      with:
        java-version: '8'
```

to create a `tmp` directory and set it in the subsequent steps. 

**Question:** 

- Do we need it here too?

rzo1 (approved):

cesarhernandezgt:
> On the `master`, we define an additional step
> 
> ```
>     steps:
>     - uses: actions/checkout@v2
>     - name: Set up JDK 1.8
>       uses: actions/setup-java@v1
>       with:
>         java-version: '8'
> ```
> 
> to create a `tmp` directory and set it in the subsequent steps as `java.io.tmpdir`.
> 
> **Question:**
> 
> * Do we need it here too?
> * Should we align the `master` workflow with this changes too? Separating the steps seems reasonable to me.

Really good questions @rzo1 .
The `tmp` directory in the github action I created on master was basically inherited form what I saw on buildbot: https://ci.apache.org/builders/tomee-trunk-ubuntu-jvm8/builds/1426

> * Do we need it here too?
I think we don't need it. My understanding is this was required in buildbot but now we are not using it on Jenkins and I don't think we need it on Github Actions.

> * Should we align the `master` workflow with this changes too? 
Yes, I think we should remove this from master: https://github.com/apache/tomee/blob/master/.github/workflows/main.yml#L23-L24

> Separating the steps seems reasonable to me.
I like it too, I'm familiar with Jenkins pipelines, and having the ability to see specific step when something went wrong is easier than trying to search within a unique file with a lot of data. To implement this approach on master will require to double-check the result we currently have when executing the profile created a couple of weeks ago that has been useful for TCK related work: -Pstyle,rat

cesarhernandezgt:
Hi @rzo1 ,

As a follow up on this conversation, I opened https://github.com/apache/tomee/pull/758 to remove the tmp folder creation on master.  I tried to separate also the checkstyle, pmd and rat into individual steps but currently `mvn checkstyle:check -Pstyle` throws 288 Checkstyle violations because on master all the check seems to be now reviewed under `-Pstyle,rat`. 

As for my next steps, I want to jump back to Jenkins CI to make sure we also have PR feedback from Jenkins because it seems the PR job is broken again https://ci-builds.apache.org/job/Tomee/job/Pull%20Requests/

PR #759, opened 2021-01-27
TOMEE-2966: Provide a pure JUnit5 OpenEJB Extension

rzo1:
# What does this PR do?

This PR aims to provide a first **JUnit 5** OpenEJB Extension to be used within JUnit 5 tests. It comes without the transient compile dependency to **JUnit 4** in **openejb-junit**.

The idea originates from a :coffee: discussion with @mawiesne in one of our research projects.

## Details

- Implements Junit 5 extensions (based on **openejb-junit** but without compile dependency to JUnit 4)  to run OpenEJB-powered tests (written in the legacy way without application composer or arquillian) in a new module **openejb-junit5-backward**.
- Adds some custom annotations to wrap "ExtendWith"
- Added (existing but migrated) tests to the new module.

# References

- https://issues.apache.org/jira/projects/TOMEE/issues/TOMEE-2966

### DIFF
--- a/container/openejb-junit/src/main/java/org/apache/openejb/junit/context/OpenEjbTestContext.java
+++ b/container/openejb-junit/src/main/java/org/apache/openejb/junit/context/OpenEjbTestContext.java
@@ -136,7 +136,9 @@ public class OpenEjbTestContext implements TestContext {
     @Override
     public void close() {
         try {
-            initialContext.close();
+            if(initialContext != null) {
+                initialContext.close();
+            }
         } catch (final NamingException e) {
             // ignored
         }
--- a/container/openejb-junit5-backward/LICENSE
+++ b/container/openejb-junit5-backward/LICENSE
@@ -1,0 +1,417 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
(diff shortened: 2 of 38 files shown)


rmannibucau:
Hi @rzo1 , from a quick review I got these points/questions but since I'm using way less TomEE it can not be relevant:

1. junit-jupiter-api should be in scope provided to let user fully decide about the version (whatever transitive dep tree he has)
2. you likely want @ExtendWtih to be replaced with @WithOpenEJB or whatever alias annotation you want
3. This part of our testing is very rarely used so wonder if it shouldn't be skipped for junit5 in favor of application composer, EJBContainer and tomee embedded flavors which are likely more mainstream and require less boilerplate code and does not use the legacy InitialContext to start the container.

Hope it helps a bit

rzo1:
Hi @rmannibucau ,

thanks for the super fast feedback.

1. I agree. 
2. This is a good idea! 
3. This is true, but we  have some users who use it in this (legacy) way, who want to migrate in a soft way without too much pain :)

rmannibucau:
@rzo1 thinking out loud but if 3 is literally 'it is there we don't want to invest" maybe do a "openejb-junit5-backward" (name is aweful but I know awesome people to find name in the project ;)) which keep using openajb-junit - just as API - and just add the junit5 glue code excluding junit artifact. This seems matching more exactly this description and enables the project to make this part slowly dying to align the maintenance on the usage?

rzo1:
@rmannibucau Yup. It is indeed some sort of backward compatibility layer to enable users, who are using the legacy way, to easily switch to JUnit 5. The way via exclusion and using the existing module as is, is a good idea. This should reduce maintenance efforts and reduces complexity (one new module versus two ...).

exabrial:
Love this! We converted to JUnit5 awhile back

PR #763, opened 2021-02-23
TOMEE-2968: Postgres connection error when a password contains "}"

rzo1:
# What does this PR do?

- Adds unit tests to reproduce TOMEE-2968 (also affects 7.0.x and 7.1.x)
- Single opening curly braces are stripped away in property values as we treat them as suffix placeholders (even in the case, we did not have an opening curly brace). PR proposes a related fix.

# References

- https://issues.apache.org/jira/browse/TOMEE-2968
- https://markmail.org/message/vxmm7t5sl33zr4wq

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/util/PropertyPlaceHolderHelper.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/util/PropertyPlaceHolderHelper.java
@@ -31,17 +31,30 @@ import java.util.Properties;
 public final class PropertyPlaceHolderHelper {
     private static final String PREFIX = "${";
     private static final String SUFFIX = "}";
+    private static final String VALUE_DELIMITER;
+
     private static final Properties CACHE = new Properties();
 
-    private static final PropertiesLookup RESOLVER = new PropertiesLookup();
-    public static final StrSubstitutor SUBSTITUTOR = new StrSubstitutor(RESOLVER);
+    private static final PropertiesLookup RESOLVER_TO_NULL_IF_MISSING = new PropertiesLookup(false);
+    private static final PropertiesLookup RESOLVER_TO_KEY_IF_MISSING = new PropertiesLookup(true);
+
+    private static final StrSubstitutor DEFAULT_SUBSTITUTOR = new StrSubstitutor(RESOLVER_TO_KEY_IF_MISSING);
+
+    private static final StrSubstitutor VALUE_DELIMITER_SUBSTITUTOR = new StrSubstitutor(RESOLVER_TO_NULL_IF_MISSING);
 
     static {
-        SUBSTITUTOR.setEnableSubstitutionInVariables(true);
-        SUBSTITUTOR.setValueDelimiter(JavaSecurityManagers.getSystemProperty("openejb.placehodler.delimiter", ":-")); // default one of [lang3]
+        VALUE_DELIMITER = JavaSecurityManagers.getSystemProperty("openejb.placehodler.delimiter", ":-"); // default one of [lang3]
+
+        DEFAULT_SUBSTITUTOR.setEnableSubstitutionInVariables(true);
+        DEFAULT_SUBSTITUTOR.setValueDelimiter(VALUE_DELIMITER);
+
+        VALUE_DELIMITER_SUBSTITUTOR.setPreserveEscapes(true);
+        VALUE_DELIMITER_SUBSTITUTOR.setEnableSubstitutionInVariables(true);
+        VALUE_DELIMITER_SUBSTITUTOR.setValueDelimiter(VALUE_DELIMITER);
     }
 
-    public static final String CIPHER_PREFIX = "cipher:";
+    private static final String CIPHER_PREFIX = "cipher:";
+    private static final String JAVA_PREFIX = "java:";
 
     private PropertyPlaceHolderHelper() {
         // no-op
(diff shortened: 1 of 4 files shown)


rmannibucau:
Hi @rzo1 , why not just escaping the character (think it is with '$' IIRC)?

rzo1:
> Hi @rzo1 , why not just escaping the character (think it is with '$' IIRC)?

Thought about it - current code will simply replace the closing bracket (even if we escape it). 

Escaping with **$** is only done for the opening version, so atm it boils down to PREFIX="${" and SUFFIX="}" to support property substitutions like "${foo}---${bar}". 

The question is, if we should replace closing / opening brackets by empty string in the first place, if the suffix/prefix environment is not complete / closed.

rmannibucau:
if incomplete noop should be done

rzo1:
> if incomplete noop should be done

Agreed. 

This behaviour is implemented in this PR. 

Before: } was replaced regardless of complete or not.

rmannibucau:
@rzo1 so maybe add a test mixing both with `${foo}}`, `$${{foo}}`, `${${bar}}`, `${$${bar}}` and so on, this is what this PR can wrongly handle (ie both at the same time and nested case).

rzo1:
> @rzo1 so maybe add a test mixing both with `${foo}}`, `$${{foo}}`, `${${bar}}`, `${$${bar}}` and so on, this is what this PR can wrongly handle (ie both at the same time and nested case).

Thanks - I added some more tests mixing different cases / types of placeholders including the escape mechanism such as `$$` to skip placeholder substitution.

rmannibucau:
@rzo1 :+1: , if you fix the regression of the default interpolation (noValueFound test and nested cases) it is good for me.

rzo1:
> @rzo1  , if you fix the regression of the default interpolation (noValueFound test and nested cases) it is good for me.

To see, if I get you right:

In case, no key is defined, we should resolve to the name of the key rather than doing a noop, ie `${v}` should resolve to `v` if no key for `v` is defined?

rmannibucau:
@rzo1 yes, this enables for simple configuration to use some default as override key (quite useful for hosts for example) but also to nested keys more readably than with defaults - and was from the time defaults were not supported) so `${${foo}.bar}` would evaluate foo.bar if foo is missing (else <foo value>.bar). Side note: if it helps, fork commons-lang3 code, openejb-core shoudln't depend on it anyway and code is quite small - https://github.com/apache/openwebbeans-meecrowave/blob/master/meecrowave-core/src/main/java/org/apache/meecrowave/lang/Substitutor.java.

rzo1:
> @rzo1 yes, this enables for simple configuration to use some default as override key (quite useful for hosts for example) but also to nested keys more readably than with defaults - and was from the time defaults were not supported) so `${${foo}.bar}` would evaluate foo.bar if foo is missing (else .bar). Side note: if it helps, fork commons-lang3 code, openejb-core shoudln't depend on it anyway and code is quite small - https://github.com/apache/openwebbeans-meecrowave/blob/master/meecrowave-core/src/main/java/org/apache/meecrowave/lang/Substitutor.java.

Regression should be fixed now - at least, if I understood it correctly.

`openejb-core` has some more dependencies towards `lang3`

- org.apache.commons.lang3.StringUtils
- org.apache.commons.lang3.text.StrLookup;
- org.apache.commons.lang3.text.StrSubstitutor; (alternative: Substitutor from `text` - might be worth replacing it as it is deprecated...)
- org.apache.commons.lang3.builder.ToStringBuilder
- org.apache.commons.lang3.ClassUtils
 - org.apache.commons.lang3.tuple.ImmutablePair

rmannibucau:
@rzo1 doesnt `if(raw.startsWith(ESCAPE_SEQUENCE)) return decryptIfNeeded(raw, acceptCharArray);` skips too much? what about `$${foo} ${bar}`?

rzo1:
Hmm yes - you are right. Will fix it soon.

rzo1:
I think, we need to discuss, which one is the expected behaviour for `$${foo}`:

1.  `$${foo}` ->  `$${foo}` (noop)
2.  `$${foo}` ->` `${foo}` (escaped sequence, no replacement of placeholder (as mentioned in the docs of `StrSubstitutor`))
3. `$${foo}` -> `foo`

cesarhernandezgt:
I think it will be  2. $${foo} ->${foo}
I'm catching up with this ticket. 

As a side note, Jenkins Job was green, then a new build hanged the job so I cancel and after your upcoming push, I hope the status on the PR gets in sync.

rzo1:
I think, that I have fixed the regression introduced while working on the escaped sequences. 

The code should now reduce escaped placeholders according to **2.**

rzo1:
@cesarhernandezgt Can we manually re-trigger the Jenkins build or does it a scheduled build. I think the failing tests are unrelated.

cesarhernandezgt:
@rzo1 I retrigger the build here: https://ci-builds.apache.org/job/Tomee/job/pull-request/81/

rzo1:
@cesarhernandezgt Thanks - the build looks good now 🙃

cesarhernandezgt (approved):

container/openejb-core/src/main/java/org/apache/openejb/util/PropertyPlaceHolderHelper.java
```diff
@@ -53,33 +55,34 @@ public static void reset() {
...
 
         String value = SUBSTITUTOR.replace(raw);
+
         if (!value.equals(raw) && value.startsWith("java:")) {
             value = value.substring(5);
         }
-        return decryptIfNeeded(value.replace(PREFIX, "").replace(SUFFIX, ""), true);
+
+        // no key defined and no escape sequence found -> substitute to key instead
+        if (!(raw.contains(ESCAPE_SEQUENCE) || value.contains(ESCAPE_SEQUENCE))
```
rmannibucau, line 80: I'm not sure you can implement this with "contains" checks (only t"here is none of these chars" test can be reliable) since you need the presence or not of the char + the position between the present ones to have a reliable test otherwise you can fall back in the same issue you originall hit which is to call "replace" where it is not needed.

Test are missing all these weirdo cases but since it is more or less the one you hit I guess it can happen.

A probably trivial fix is to extend PropertiesLookup to handle the fallback replacement (instead of null it returns the key) and just always call the substitutor if there is at least start+end chars. Worse case the substitutor will already do nothing.

Wdyt?

rzo1: Thx @rmannibucau  for the follow up!

Yes - testing with `contains` is a simplification and neglects some weirdo cases... and yes: it would require char, position + offsets to properly test and replace the closing and opening brackets. 

Thus, I like the idea of using `PropertiesLookup#lookup` and returning the key there. The parsing "magic" (char, position, offsets) is already done in the substitutor, so I guess fallbacking there is a good option. 

I will take a look at it.

rzo1: @rmannibucau 

Wouldn't returning the `key` instead of `null` in `PropertiesLookup#lookup` breaking the `:-` syntax? 

From my understanding, `:-` is used for some sort of inlined "if else", i.e. "if foo is not defined, use bar instead" - is this assumption correct? Or maybe I didn't understood the substition process for `:-` correctly? :)

For example:

- "${foo:-bar}" -> `bar` (in the original implementation (returning `null`); `foo` is undefined)
- "${foo:-bar}" -> `foo` (if we return the `key` instead of `null`, the second part via  `:-` is not evaluated as we have a match for the given key (which is the key itself))

rmannibucau (approved):

PR #767, opened 2021-03-09
TOMEE-2977 - Provide a ApplicationComposer JUnit 5 Extension

rzo1:
# What does this PR do?

- Adds a new module `openejb-junit5`
- Adds a JUnit 5 extension, which internally uses `ApplicationComposers` for (pure) JUnit 5 testing. 
- Adds some unit tests (based on existing `ApplicationComposer` tests) written in JUnit 5
- Provides an additional JUnit 5 example (based on `application-composer`) 
- Upgrades Maven surefire to `3.0.0-M5` ( is required to run both variants of unit tests without workarounds out of the box)

# References 

- https://issues.apache.org/jira/projects/TOMEE/issues/TOMEE-2977

### DIFF
--- a/container/openejb-core/pom.xml
+++ b/container/openejb-core/pom.xml
@@ -313,10 +313,10 @@
           </execution>
         </executions>
       </plugin>
+
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
-        <version>${surefire.version}</version>
         <configuration>
           <childDelegation>true</childDelegation>
           <reuseForks>false</reuseForks>
--- a/container/openejb-junit5/pom.xml
+++ b/container/openejb-junit5/pom.xml
@@ -1,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one or more
+    contributor license agreements.  See the NOTICE file distributed with
+    this work for additional information regarding copyright ownership.
+    The ASF licenses this file to You under the Apache License, Version 2.0
+    (the "License"); you may not use this file except in compliance with
+    the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>container</artifactId>
+        <groupId>org.apache.tomee</groupId>
+        <version>8.0.7-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>openejb-junit5</artifactId>
+    <packaging>jar</packaging>
+    <name>TomEE :: Container :: JUnit 5</name>
+
+    <properties>
+        <tomee.build.name>${project.groupId}.container.junit5</tomee.build.name>
+        <netbeans.hint.license>openejb</netbeans.hint.license>
+        <jacocoArgLine/>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>openejb-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.openejb.itests</groupId>
+            <artifactId>failover-ejb</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <childDelegation>true</childDelegation>
+                    <reuseForks>false</reuseForks>
+                    <forkCount>1</forkCount>
+                    <argLine>
+                        ${jacocoArgLine}
+                        -enableassertions
+                    </argLine>
+                    <workingDirectory>${project.basedir}/target</workingDirectory>
+                    <systemPropertyVariables>
+                        <openejb.home>${project.basedir}/target/test-classes</openejb.home>
+                    </systemPropertyVariables>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>default-test</id>
+                        <phase>test</phase>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <configuration>
+                            <excludes>
+                                <exclude>org/apache/openejb/junit5/SingleAppComposerTest</exclude>
+                                <exclude>org/apache/openejb/junit5/SingleAppComposerJVMTest</exclude>
+                            </excludes>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>per-jvm-tests</id>
+                        <phase>test</phase>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <configuration>
+                            <forkCount>0</forkCount>
+                            <includes>
+                                <include>org/apache/openejb/junit5/SingleAppComposerTest</include>
+                                <include>org/apache/openejb/junit5/SingleAppComposerJVMTest</include>
+                            </includes>
+                            <argLine>
+                                ${jacocoArgLine}
+                                -enableassertions
+                                -Djunit.jupiter.testclass.order.default=org.apache.openejb.junit5.order.AppComposerTestClassOrderer
+                                -Dtomee.application-composer.application=org.apache.openejb.junit5.app.MyApp
+                            </argLine>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/container/openejb-junit5/src/main/java/org/apache/openejb/junit5/AfterReleaserBase.java
+++ b/container/openejb-junit5/src/main/java/org/apache/openejb/junit5/AfterReleaserBase.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.openejb.junit5;
+
+import org.apache.openejb.testing.ApplicationComposers;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public abstract class AfterReleaserBase extends ApplicationComposerExtensionBase {
+
+    private final ExtensionContext.Namespace namespace;
+
+    AfterReleaserBase(ExtensionContext.Namespace namespace) {
+        this.namespace = namespace;
+    }
+
+    void run(final ExtensionContext extensionContext) throws Exception {
+        extensionContext.getStore(namespace).get(ApplicationComposers.class, ApplicationComposers.class).after();
+    }
+}
--- a/container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerExtension.java
+++ b/container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerExtension.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.openejb.junit5;
+
+import org.apache.openejb.OpenEJBRuntimeException;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public class ApplicationComposerExtension extends ApplicationComposerExtensionBase implements BeforeAllCallback, AfterAllCallback, BeforeEachCallback, AfterEachCallback {
+
+    private static final ExtensionContext.Namespace NAMESPACE = ExtensionContext.Namespace.create(ApplicationComposerExtension.class.getName());
+
+    private final Object[] modules;
+
+    public ApplicationComposerExtension() {
+        this((Object[]) null);
+    }
+
+    public ApplicationComposerExtension(Object... modules) {
+        this.modules = modules;
+    }
+
+    @Override
+    public void beforeAll(ExtensionContext context) throws Exception {
+
+        if (isPerJvm(context)) {
+            if(this.modules != null && this.modules.length > 0) {
+                throw new IllegalArgumentException("Modules are not supported in PER_JVM mode. Set 'tomee.application-composer.application' system property.");
+            }
+            context.getStore(NAMESPACE).put(ApplicationComposerPerXYExtensionBase.class, new ApplicationComposerPerJVMExtension());
+        } else if (isPerAll(context)) {
+            context.getStore(NAMESPACE).put(ApplicationComposerPerXYExtensionBase.class, new ApplicationComposerPerAllExtension(this.modules));
+        } else if (isPerEach(context)) {
+            context.getStore(NAMESPACE).put(ApplicationComposerPerXYExtensionBase.class, new ApplicationComposerPerEachExtension(this.modules));
+        } else if (isPerDefault(context)) {
+            context.getStore(NAMESPACE).put(ApplicationComposerPerXYExtensionBase.class, new ApplicationComposerPerDefaultExtension(this.modules));
+        } else {
+            throw new OpenEJBRuntimeException("No ExtensionMode is present.");
+        }
+
+        context.getStore(NAMESPACE).get(ApplicationComposerPerXYExtensionBase.class, ApplicationComposerPerXYExtensionBase.class).beforeAll(context);
+    }
+
+    @Override
+    public void afterAll(ExtensionContext context) throws Exception {
+        ApplicationComposerPerXYExtensionBase delegate = context.getStore(NAMESPACE).get(ApplicationComposerPerXYExtensionBase.class, ApplicationComposerPerXYExtensionBase.class);
+        if(delegate != null) {
+            delegate.afterAll(context);
+        }
+    }
+
+    @Override
+    public void beforeEach(ExtensionContext context) throws Exception {
+        ApplicationComposerPerXYExtensionBase delegate = context.getStore(NAMESPACE).get(ApplicationComposerPerXYExtensionBase.class, ApplicationComposerPerXYExtensionBase.class);
+        if(delegate != null) {
+            delegate.beforeEach(context);
+        }
+    }
+
+    @Override
+    public void afterEach(ExtensionContext context) throws Exception {
+        ApplicationComposerPerXYExtensionBase delegate = context.getStore(NAMESPACE).get(ApplicationComposerPerXYExtensionBase.class, ApplicationComposerPerXYExtensionBase.class);
+        if(delegate != null) {
+            delegate.afterEach(context);
+        }
+    }
+
+}
--- a/container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerExtensionBase.java
+++ b/container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerExtensionBase.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.openejb.junit5;
+
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.platform.commons.util.AnnotationUtils;
+
+import java.util.Optional;
+
+public abstract class ApplicationComposerExtensionBase {
+
+    boolean isPerClassLifecycle(final ExtensionContext context) {
+        return isPerTestInstanceLifecycle(context, TestInstance.Lifecycle.PER_CLASS);
+    }
+
+    boolean isPerMethodLifecycle(final ExtensionContext context) {
+        return isPerTestInstanceLifecycle(context, TestInstance.Lifecycle.PER_METHOD);
+    }
+
+    boolean isPerTestInstanceLifecycle(final ExtensionContext context, TestInstance.Lifecycle lifecycle) {
+        return context.getTestInstanceLifecycle()
+                .map(it -> it.equals(lifecycle))
+                .orElse(false);
+    }
+
+    protected boolean isPerEach(final ExtensionContext context) {
+        return checkMode(context, ExtensionMode.PER_EACH);
+    }
+
+    boolean isPerAll(final ExtensionContext context) {
+        return checkMode(context, ExtensionMode.PER_ALL);
+    }
+
+    boolean isPerJvm(final ExtensionContext context) {
+        return checkMode(context, ExtensionMode.PER_JVM);
+    }
+
+    boolean isPerDefault(final ExtensionContext context) {
+        return checkMode(context, ExtensionMode.AUTO);
+    }
+
+    boolean checkMode(final ExtensionContext context, ExtensionMode extensionMode ) {
+       return extensionMode == getModeFromAnnotation(context);
+    }
+
+    ExtensionMode getModeFromAnnotation(final ExtensionContext context) {
+        return context.getTestClass()
+                .flatMap(test -> AnnotationUtils.findAnnotation(test, RunWithApplicationComposer.class))
+                .map(RunWithApplicationComposer::mode)
+                .orElse(ExtensionMode.AUTO);
+    }
+}
--- a/container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerPerAllExtension.java
+++ b/container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerPerAllExtension.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.openejb.junit5;
+
+import org.apache.openejb.util.LogCategory;
+import org.apache.openejb.util.Logger;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public class ApplicationComposerPerAllExtension extends ApplicationComposerPerXYExtensionBase implements BeforeAllCallback, AfterAllCallback, BeforeEachCallback, AfterEachCallback {
+
+    private static final Logger logger = Logger.getInstance(LogCategory.OPENEJB, ApplicationComposerPerAllExtension.class);
+
+    public ApplicationComposerPerAllExtension() {
+        this((Object[]) null);
+    }
+
+    public ApplicationComposerPerAllExtension(Object... modules) {
+        super(modules);
+    }
+
+    @Override
+    protected void validate(ExtensionContext context) {
+        super.validate(context);
+        if (isPerAll(context) && isPerMethodLifecycle(context)) {
+            logger.info("Running PER_ALL in combination with TestInstance.Lifecycle.PER_METHOD.");
+            logger.info("Please note, there are some limitations (N = amount of test methods):");
+            logger.info("N = 1: Will work as expected.");
+            logger.info("N > 1: Injections are lost after the first test method was executed.");
+            logger.info("N > 1: Use CDI.current(), InitialContext or pure (http) client to implement the test.");
+        }
+    }
+
+    @Override
+    public void beforeAll(ExtensionContext context) throws Exception {
+        super.beforeAll(context);
+        doInit(context);
+        doStart(context);
+        doInject(context);
+        addAfterAllReleaser(context);
+    }
+}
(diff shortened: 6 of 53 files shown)


container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.openejb.junit.jupiter;
+
+import org.apache.openejb.OpenEJBRuntimeException;
+import org.apache.openejb.testing.ApplicationComposers;
+import org.junit.jupiter.api.extension.*;
+
+public class ApplicationComposerExtension implements BeforeEachCallback, AfterEachCallback {
```
rmannibucau, line 23: A few open questions:

1. isnt it beforeall/afterall callbacks? (way faster)
2. shouldn't "delegate" be stored in ExtensionContext to ensure it is well scoped and not limited to this particular threading+scope which is one of the multiple junit5 cases
3. What about other test instances you can get from the context? should they contribute to the app model?
4. Should it get a "Single"/mono extension (started once per jvm as tomee embedded), it is generally accurate for apps written this way
5. (very minor) I wouldn't mention extension in the annotation "ApplicatioComposerExtension" name (maybe `@WithApplicationComposer` or alike)

hope it makes sense

rzo1: Wow. Super fast :+1:

1. I thought about it: Doesn't `ApplicationComposers` require a concrete test instance to inject values and stuff in `delegate.before(...)`?` The required test instance is - afaik - not available when `beforeAll(...)` is called. However, the creation of the `delegate` might be conducted in a `beforeAll(...) call.
2. This is a good idea. I think, it can be done via `context.getRoot().getStore(...)`  the `ExtensionContext`.
3. Honestly, I did not thought about this use-case. We may get additional test instances via `context.getRequiredTestInstances().getAllInstances()`. Wdyt?
4. Should be very similar to `SingleApplicationComposerRunner` - I will take a look at it.
5. Good idea. Will change it to `RunWith...` to be inline with the namings in `openejb-junit5-backward`.

rzo1: Thx. I updated things related to 2. and 3.

I have to think about 1. (because I think, that I didnt get it) and read / study the documentation related to `DynamicTests` in JUnit 5. Do you have an example of "splitting test definitions into multiple classes"? 

Might get you wrong, but did you refer to  `Test Instance Lifecycles` when talking about `@ApplicationComposer(mode = [PER_JVM|PER_ALL|PER_EACH])`  ? 

So we get smth like this, if the mode is specified?

- PER_JVM --> `SingleApplicationComposerExtension` (like TomEE embedded?)
- PER_ALL --> `@TestInstance(Lifecycle.PER_CLASS)` //  new test instance will be created once per test class.
- PER_EACH --> `@TestInstance(Lifecycle.PER_METHOD)` // a new test instance will be created for each test method, test factory method, or test template method (default in JUnit 4)

rzo1: Thanks for the details! 

> The split of an app composer app in multiple class is actually what you use ;). The constructor of ApplicationComposers takes N objects, just pass all test classes if you want a single deployment for all tests and enable each test to deploy part of the app.

Afaik the extension context only provides the related test class. So it would be something like

- Get inner classes from the test class
- Create instances of these inner test classes
- Supply the test class as well as the N instances of the inner test classes to  `ApplicationComposers` (like it is done in `AppComposerAsJUnitRuleWithReusableModulesTest` via JUnit 4 rules)  ?

> About the lifecycle it is not 1-1 with test instance lifecycle but the same idea for the container lifecycle (but you can have one test instance per test and one container instance for all tests of a single class when you want to reset another extension data/field instances for example). We can align on the default if nothing is set (so add to the enum an AUTO mode ;)) but must be configurable.

Thanks!  I think, I am now getting a better understanding of the idea. I also checked the JUnit 5 extension impl of meecrowave, which was really helpful.

rzo1: I made some progress based on your feedback, but I think it is still heavily **WIP** ;)  It should now honor the test instance lifecycle. The `mode` thing is missing yet. I have to think about the impl for the proposed `mode` for controlling the container instances created during tests.

rzo1: The current impl relies on the test's life cycle to start a container or perform related injections.

Atm the impl contains annotations and extensions for

- `RunWithSingleApplicationComposer`
- `RunWithApplicationComposer`

as well as a 

- `TomEEEmbeddedExtension`

From my understanding, the proposed `mode` is solely responsible to configure how (or when) the underlying container is started? I.e. one container to run all tests within a single test class (PER_ALL), one container for each test method of one test class (PER_EACH) and one container to run multiple test classes (PER_JVM), right?  Is this the same idea, which is currently conducted through different JUnit4 runners? 

From my understanding of the current JUnit implementation, the `ApplicationComposer` seems to be something similar to the PER_ALL and the `EmbeddedTomEERunner` / `SingleApplicationComposer` to the PER_JVM; I think PER_EACH (test method) isn't implemented in the JUnit 4 runners yet.

Probably I am missing something but I think, I do not have the whole process or related use-cases in mind ;) - Maybe you , @rmannibucau , can add some more ideas / thoughts on the `mode` idea?

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtensionBase.java
```diff
@@ -26,4 +27,36 @@ boolean isPerClass(final ExtensionContext context) {
...
+    }
+
+    boolean checkMode(final ExtensionContext context, ExtensionMode extensionMode ) {
+       return extensionMode == getModeFromAnnotation(context);
+    }
+
+    ExtensionMode getModeFromAnnotation(final ExtensionContext context) {
+        if (context.getTestClass().isPresent()) {
+
+            RunWithApplicationComposer a = context.getTestClass().get().getAnnotation(RunWithApplicationComposer.class);
```
rmannibucau, line 54: AnnotationUtils.findAnnotation ;) (will support junit5 stereotypes otherwise you loose it by pure reflection)

rzo1: Thx. Changed it.

container/openejb-core/pom.xml
```diff
@@ -580,6 +580,22 @@
       <groupId>org.apache.xbean</groupId>
       <artifactId>xbean-bundleutils</artifactId>
     </dependency>
+    <dependency>
```
rmannibucau, line 583: open point: should it go in a new openejb-junit5 module? Having testing stack in openejb-core is an inheritance but bothering at some point so maybe time to clean it up? wdyt?

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+
+public class ApplicationComposerExtension extends ApplicationComposerExtensionBase implements BeforeAllCallback, AfterAllCallback, BeforeEachCallback, AfterEachCallback {
+
+    private static final ExtensionContext.Namespace NAMESPACE = ExtensionContext.Namespace.create(ApplicationComposerExtension.class.getName());
+    private static final SingleApplicationComposerBase BASE = new SingleApplicationComposerBase();
+
+    @Override
+    public void beforeAll(ExtensionContext context) throws Exception {
+
+        if (isPerJvm(context)) {
```
rmannibucau, line 37: If you have per jvm and per test/class at the same time what happens? Options I see:

1. test if per jvm is started and fail perall/pereach cases
2. make it work (we did in meecrowave, it is mainly a matter of having a JVM singleton/lock to switch the tccl for the bootstrap time + a few singleton fixes)

I'd start with 1 keeping 2 in mind (since it requires more work) so something like 'if !JVM && BASE.isStarted() => fail', isStarted is just matter of testing one of the atomic ref of the single instance.

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+        if (isPerEach(context)) {
+            doRelease(context);
+        } else if (isPerDefault(context) && !isPerClass(context)) {
+            doRelease(context);
+        }
+    }
+
+    private void doRelease(final ExtensionContext extensionContext) throws Exception {
+        if (isPerJvm(extensionContext)) {
+            //FIXME how to release / close after last test class execution?
```
rmannibucau, line 87: should be done with the shutdown hook
this implementation does NOT work so it likely means it misses a test which should fail (like chaining 2 test classes and ensuring it uses the same container)

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+public class ApplicationComposerExtension extends ApplicationComposerExtensionBase implements BeforeAllCallback, AfterAllCallback, BeforeEachCallback, AfterEachCallback {
+
+    private static final ExtensionContext.Namespace NAMESPACE = ExtensionContext.Namespace.create(ApplicationComposerExtension.class.getName());
+    private static final SingleApplicationComposerBase BASE = new SingleApplicationComposerBase();
+
+    @Override
+    public void beforeAll(ExtensionContext context) throws Exception {
+
+        if (isPerJvm(context)) {
+            BASE.start(context.getTestClass().orElse(null));
```
rmannibucau, line 38: do you want to check there is no @Module/@Classes/... or so in the class? it would be a misuse of the API

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+        } else if (isPerAll(context)) {
+            doStart(context);
+        } else if (isPerDefault(context)) {
+            if (isPerClass(context)) {
+                doStart(context);
+            }
+        }
+
+        if (isPerClass(context)) {
```
rmannibucau, line 47: not sure it is intended but previous code makes it possible to not have started anything there and if so it must not inject anything
I suspect not starting at all is a bug so maybe rework the if/else flow to ensure we start?

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+            doInject(context);
+        }
+    }
+
+    @Override
+    public void afterAll(ExtensionContext context) throws Exception {
+        if (isPerJvm(context) || isPerAll(context)) {
```
rmannibucau, line 54: perjvm => do not release?

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+            doRelease(context);
+        } else if (isPerDefault(context) && isPerClass(context)) {
```
rmannibucau, line 56: my 2 cts would be to create an AfterAllReleaser {void run()} and put it in the context in beforeAll next to the related start logic.
Then afterXXX is just context.getNamespace(...).getStore().get(AfterAllReleaser.class, AfterAllReleaser.class).ifPresent(AfterAllReleaser::run);

Same can apply to afterEach.

Big advantage is you don't duplicate all tests, it is symmetric by construction.

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+            doRelease(context);
+        }
+    }
+
+
+    @Override
+    public void beforeEach(ExtensionContext context) {
+
+        if (isPerEach(context)) {
```
rmannibucau, line 65: Think I'd try to merge tests which would also enable to move the doInject next the doStart which would help to simplify the flow.

Open question: if beforeAll adds something specific in the store then if it is not in the store there it means it is not "all" or "per jvm" so you start and inject there no? Isn't it easier to just test an "all" marker for each impl?

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        if (!isPerClass(context)) {
+            doInject(context);
+        }
+    }
+
+    @Override
+    public void afterEach(ExtensionContext context) throws Exception {
+        if (isPerEach(context)) {
```
rmannibucau, line 78: I'd use the same AfterEachReleaser hack (side note: name is not great but you get the idea ;)).

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+            //FIXME how to release / close after last test class execution?
+            BASE.close();
+        } else {
+            extensionContext.getStore(NAMESPACE).get(ApplicationComposers.class, ApplicationComposers.class).after();
+        }
+    }
+
+    private void doStart(final ExtensionContext extensionContext) {
+
+        Optional<Class<?>> oClazz = extensionContext.getTestClass();
```
rmannibucau, line 96: > final Class<?> oClazz =  extensionContext.getTestClass().orElseThrow(() ->new OpenEJBRuntimeException("Could not get test class from the given extension context."));

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+            throw new OpenEJBRuntimeException("Could not get test class from the given extension context.");
+        }
+
+        extensionContext.getStore(NAMESPACE).put(ApplicationComposers.class,
+                new ApplicationComposers(oClazz.get(), getAdditionalModules(oClazz.get())));
+
+    }
+
+    private void doInject(final ExtensionContext extensionContext) {
+        Optional<TestInstances> oTestInstances = extensionContext.getTestInstances();
```
rmannibucau, line 108: orElseThrow ;)

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+                    throw new OpenEJBRuntimeException(e);
+                }
+            });
+        } else {
+            ApplicationComposers delegate = extensionContext.getStore(NAMESPACE)
+                    .get(ApplicationComposers.class, ApplicationComposers.class);
+
+            testInstances.forEach(t -> {
+                try {
+                    delegate.before(t);
```
rmannibucau, line 130: this does not injects but starts a container so this is not supposed to work - or at least as expected, maybe add a test to validate this case

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+                } catch (Exception e) {
+                    throw new OpenEJBRuntimeException(e);
+                }
+            });
+        }
+    }
+
+    private Object[] getAdditionalModules(final Class<?> clazz) {
+        return Arrays.stream(clazz.getClasses())
```
rmannibucau, line 139: this likely do not do what is expected and can be random (thinking to @Nested tests of junit5).

ApplicationComposerRule shows the idea: you pass the list of additional modules.

In junit5 it can be:

> @RegisterExtension ApplicationComposerExtension ext = new ApplicationComposerExtension(new MyApp());

it enables to define the app outside the test and/or to compose subapps definitions (like a MyStack() + finish the spec in the test)

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,155 @@
...
+                .map(this::newInstance)
+                .filter(Objects::nonNull)
+                .toArray();
+    }
+
+    private Object newInstance(final Class<?> clazz) {
+        try {
+            return clazz.newInstance();
```
rmannibucau, line 147: > clazz.getConstructor().newInstance()

blame java, not me ;)

(only relevant if you decide to still use nested classes which is not needed regarding previous comment)

examples/junit5-application-composer/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+        <version>3.5.1</version>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.22.1</version>
```
rmannibucau, line 45: use 3.0.0-M5 please, previous versions have all pitfalls with junit5

examples/junit5-application-composer/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.tomee</groupId>
+      <artifactId>javaee-api</artifactId>
+      <version>[8.0,)</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
```
rmannibucau, line 84: <artifactId>junit-jupiter</artifactId> as single junit5 dep

pom.xml
```diff
@@ -97,8 +97,7 @@
     <tomee.version>${project.version}</tomee.version>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
-    <surefire.version>2.21.0</surefire.version>
-    <surefire.junit5.version>3.0.0-M5</surefire.junit5.version>
+    <surefire.version>2.22.1</surefire.version>
```
rmannibucau, line 100: don't downgrade, as mentionned iit can fake it works with junit5 whereas it actually does not

tomee/tomee-embedded/src/main/java/org/apache/tomee/embedded/junit/jupiter/TomEEEmbeddedExtension.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.openejb.OpenEJBRuntimeException;
+import org.apache.openejb.junit.jupiter.ApplicationComposerExtensionBase;
+import org.apache.tomee.embedded.junit.TomEEEmbeddedBase;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.extension.*;
+
+import java.util.List;
+import java.util.Optional;
+
+public class TomEEEmbeddedExtension extends ApplicationComposerExtensionBase implements BeforeAllCallback, AfterAllCallback, BeforeEachCallback, AfterEachCallback {
```
rmannibucau, line 28: maybe also define the @RunWithTomEEEmbedded 

Also wonder if ApplicationComposerExtensionBase is needed, as seen in app composer case, you have beforeAll to test if you are  in JVM or per class mode, then all other tests are deduced from the content of the context store so not sure it is good to define a dependency between  both modules just to test a scope. (from an artifact/dep point of view we should be able to run tomee embedded without application composer impl, only its org.apache.openejb.testing API)

tomee/tomee-embedded/src/main/java/org/apache/tomee/embedded/junit/jupiter/TomEEEmbeddedExtension.java
```diff
@@ -0,0 +1,86 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.tomee.embedded.junit.jupiter;
+
+import org.apache.openejb.OpenEJBRuntimeException;
+import org.apache.openejb.junit.jupiter.ApplicationComposerExtensionBase;
+import org.apache.tomee.embedded.junit.TomEEEmbeddedBase;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.extension.*;
```
rmannibucau, line 23: wildcard? ;)

rmannibucau (changes requested):

tomee/tomee-embedded/src/main/java/org/apache/tomee/embedded/junit/jupiter/TomEEEmbeddedExtension.java
rzo1: Good catch. `ApplicationComposerExtensionBase` isn't needed nor is it used.

container/openejb-core/pom.xml
rzo1: I think it would be (a) cleaner and (b) could allow user's to exclude the compile dependencies towards junit4  - so +1 for `openejb-junit5`. 

Could also include the classes from `openejb-junit5-backward`, so we have all junit 5 related things in one module. wdyt?

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
rzo1: Couldn't find any documentation related to `@Application` at all. It seems, it isn't quite often used in the code base nor in the examples? 

`ApplicationComposers#validate()`will complain and fail to start any container, if there is no `@Classes` or `@Default` annotation present?

Probably, I don't get the point here?

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
rzo1: I updated the related code fragments (via `validate(context)`. I don't know exactly, if I got the idea. Can you check, @rmannibucau ?

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
rzo1: I think, I do not get the idea of the "all"-marker ? Is it about simplyfing the different conditions (which are some how reworked now but maybe subject to change) 

Maybe you can add some more context, @rmannibucau ?

rmannibucau: Yes, my starting point is that it is very error prone to test multiple times the same thing (let's ignore perf since it is simple tests). Since any "start" logic will put something (likely different) in the store, the presence can be used to test which case you are in to handle > start callbacks.
One way to see it is to have 3 impls (3 extensions) of BeforeAll+BeforeEach+AfterAll+AfterEach and put the impl in beforeAll of this extension which will then acts as a facade and select which impl to use in before all.
This way code path is unique each time and simple, only this extension beforeAll has some mode usage, all others are passthrough.

rzo1: Thx for clarification. Makes sense to me. I have updated the related code: https://github.com/apache/tomee/pull/767/commits/fc58a622b9f6aa8aa9ae6d69f91bcef11a4c11cc

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
rzo1: 1.) -> might be addressed with https://github.com/apache/tomee/pull/767/commits/6fba175b05270b9838b87a98d346da246135c4d0

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
rzo1: Okay. I think I am understanding it now. Found some information about `-Dtomee.application-composer.application`, so I guess, we should do the validation.

container/openejb-core/pom.xml
rzo1: Just moved everything into `openejb-junit5`. 

Should we also include the contents of `openejb-junit5-backward` in this new  `openejb-junit5` module? We haven't released it yet, so it wouldn't bother anyone.

Would bring a dependency towards the legacy testing stuff:

```
 <dependency>
            <groupId>${project.groupId}</groupId>
            <artifactId>openejb-junit</artifactId>
            <version>${project.version}</version>
        </dependency>
```

Might be cleaner without it? wdyt @rmannibucau ?

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
rzo1: If I get it right, the PER_JVM case takes a fully assembled `@Application` defined in a class specified via `tomee.application-composer.application`. The rational is now to check for additional annotations (modules, classes, ...), which would lead to undefined behaviour. So it requires a check / validation for the following annotations

```
org.apache.openejb.testing.Component; //??
org.apache.openejb.testing.Module;
org.apache.openejb.testing.Classes;
org.apache.openejb.testing.Default //??
org.apache.openejb.testing.Jars; // ??
//similar @Deprecated annotated annotations in *.junit.*
```
on the test class and its related methods? Anything additional, @rmannibucau ?

rmannibucau: guess you can skip @Deprecated ones, they shouldnt be used in junit5 module anyway (new code with deprecated API sounds fishy from the IDE ;)).

container/openejb-core/pom.xml
rzo1: 1. +1 -> I agree.
2. Yes. It is a totally different approach, so keeping them separated is fine for me. Users may consume them as needed.

container/openejb-core/src/main/java/org/apache/openejb/junit/jupiter/ApplicationComposerExtension.java
rzo1: Alright - just added a related validation.

rzo1:
Thanks for the review, the feedback and the comments, @rmannibucau. If you don't mind, it might be ready for a 2nd review round ... ;)

container/openejb-junit5/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+                    <forkCount>1</forkCount>
+                    <argLine>
+                        ${jacocoArgLine}
+                        -enableassertions
+                    </argLine>
+                    <workingDirectory>${project.basedir}/target</workingDirectory>
+                    <systemPropertyVariables>
+                        <openejb.home>${project.basedir}/target/test-classes</openejb.home>
+                    </systemPropertyVariables>
+                    <excludes>
```
rmannibucau, line 81: an excluded test is a test to delete ;)

joke apart, use N executions to ensure all tests run in the build.

container/openejb-junit5/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>${surefire.junit5.version}</version>
```
rmannibucau, line 68: we should have a single surefire version for the full build IMHO

container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerExtensionBase.java
```diff
@@ -0,0 +1,72 @@
...
+    boolean isPerDefault(final ExtensionContext context) {
+        return checkMode(context, ExtensionMode.AUTO);
+    }
+
+    boolean checkMode(final ExtensionContext context, ExtensionMode extensionMode ) {
+       return extensionMode == getModeFromAnnotation(context);
+    }
+
+    ExtensionMode getModeFromAnnotation(final ExtensionContext context) {
+        if (context.getTestClass().isPresent()) {
```
rmannibucau, line 62:
```
return context.getTestClass()
  .flatMap(test -> AnnotationUtils.findAnnotation(test, RunWithApplicationComposer.class))
  .map(RunWithApplicationComposer::mode)
  .orElse(ExtensionMode.AUTO);
```

Overall optional.get() should almost never be used ;)

container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerPerXYExtensionBase.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public ApplicationComposerPerXYExtensionBase() {
+        this((Object[]) null);
+    }
+
+    public ApplicationComposerPerXYExtensionBase(Object... modules) {
+        this.modules = modules;
+    }
+
+    abstract void validate(ExtensionContext context);
```
rmannibucau, line 44: I'd put it protected and not package scope, wdyt?

container/openejb-junit5/src/main/java/org/apache/openejb/junit5/AfterReleaserBase.java
```diff
@@ -0,0 +1,38 @@
...
+
+    AfterReleaserBase(ExtensionContext.Namespace namespace) {
+        this.namespace = namespace;
+    }
+
+    void run(final ExtensionContext extensionContext) throws Exception {
+        doRelease(extensionContext);
+    }
+
+    void doRelease(final ExtensionContext extensionContext) throws Exception {
```
rmannibucau, line 34: shouldn't it be merged with run? don't think it is needed to split anymore, wdyt?

container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerPerAllExtension.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public ApplicationComposerPerAllExtension(Object... modules) {
+        super(modules);
+    }
+
+    @Override
+    void validate(ExtensionContext context) {
+        if (isPerAll(context) && isPerMethodLifecycle(context)) {
+            //XXX: Might be possible to make this work - for now, we are going to throw something...
+            throw new OpenEJBRuntimeException("Cannot run PER_ALL in combination with TestInstance.Lifecycle.PER_METHOD.");
```
rmannibucau, line 40: maybe just log a warning/multiline info log, it means the injections will be lost at test > 1 but for 1 test it is fine and for N > 1 tests it works if you use CDI.current() or InitialContext or pure client (http) to impl the test.

container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerPerEachExtension.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    void validate(ExtensionContext context) {
+        //no-op
+    }
+
+    @Override
+    public void afterAll(ExtensionContext context) throws Exception {
+        addAfterEachReleaser(context);
```
rmannibucau, line 42: this one looks fishy, shouldnt it be in beforeEach?

container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerPerDefaultExtension.java
```diff
@@ -0,0 +1,59 @@
...
+    public ApplicationComposerPerDefaultExtension() {
+        this((Object[]) null);
+    }
+
+    public ApplicationComposerPerDefaultExtension(Object... modules) {
+        super(modules);
+    }
+
+    @Override
+    void validate(ExtensionContext context) {
```
rmannibucau, line 36: I suspect parent one can be empty by defaul tinstead of being abstract

container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerPerJVMExtension.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @Override
+    public void beforeEach(ExtensionContext context) throws Exception {
+        if (isPerMethodLifecycle(context)) {
+            doInject(context);
+        }
+    }
+
+    private void doInject(final ExtensionContext extensionContext) {
```
rmannibucau, line 88: this one should be the same for all cases, even global one, no?

container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public ApplicationComposerExtension(Object... modules) {
+        this.modules = modules;
+    }
+
+    @Override
+    public void beforeAll(ExtensionContext context) throws Exception {
+
+        if (!isPerJvm(context) && ApplicationComposerPerJVMExtension.isStarted()) {
```
rmannibucau, line 43: let the delegate validate it in before all? the trick will be to make the validation work when this facade is not used too so each impl must validate the combinations (or the parent do it all) so the facade does not need any validation I think

container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerExtension.java
```diff
@@ -0,0 +1,87 @@
...
+            //XXX: Future work: We might get it to work via a JVM singleton/lock, see https://github.com/apache/tomee/pull/767#discussion_r595343572
+            throw new OpenEJBRuntimeException("Cannot run PER_JVM in combination with PER_ALL, PER_EACH or AUTO");
+        }
+
+        if (isPerJvm(context)) {
+            context.getStore(NAMESPACE).put(ApplicationComposerPerXYExtensionBase.class, new ApplicationComposerPerJVMExtension());
```
rmannibucau, line 49: if modules.length > 0 => throw it is unsupported, use -Dtomee.application etc

examples/junit5-application-composer/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>3.0.0-M5</version>
+        <configuration>
+          <argLine>-Djdk.attach.allowAttachSelf</argLine>
+        </configuration>
+      </plugin>
+      <plugin>
```
rmannibucau, line 50: this one is not needed since the deployment is skipped by putting it locally IMHO

rmannibucau (changes requested):
a few details to fix/enhance (no need to re-review after since they are minor) but overall looks good to me, only big blocker is the fact to skip some tests but multiplying executions will solve it

container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerPerEachExtension.java
rzo1: it is **fishy**. good catch!

container/openejb-junit5/src/main/java/org/apache/openejb/junit5/ApplicationComposerPerJVMExtension.java
rzo1: From my understanding of the code:

`ApplicationComposers.before(...)` starts a container, deploys the app and does the injection internally (in `doStart(...)`) ? In the `PER_JVM`-case, we have this single app composer thing, which runs over the full JVM lifetime and is required to inject the `@Application` into (new) test instances via `doInject(...)`.

Can you add some more details, @rmannibucau ?

rmannibucau: injection is the same in all cases (inject with the deployed app) so code path should be the same IMHO.
it is fine to make public the enrich method of the applicationcomposers instance and store in it the bean context to be able to inject later (probably expose enrich(Object) only).

rzo1: Think it is addressed now.

rzo1:
Should now contain the proposed fixes / enhancements. 

Tests are now executed in an additional phase. Let's wait if we receive some additional feedback.

PR #776, opened 2021-03-27
TOMEE-2988: Xbean 4.18 + asm9

cocorossello:


Daniel-Dos (approved):
very good for me .

rzo1:
@cesarhernandezgt Do you have an idea, why we lost the Jenkins PR builder checks for (newer) pull requests?

arquillian/arquillian-tomee-common/pom.xml
```diff
@@ -86,6 +86,16 @@
       <artifactId>xbean-finder-shaded</artifactId>
       <version>${xbeanVersion}</version>
       <scope>provided</scope>
+      <exclusions>
```
rzo1, line 89: I am wondering:

Is this needed?  Afaik [xbean-finder-shaded](https://mvnrepository.com/artifact/org.apache.xbean/xbean-finder-shaded/4.18) only has a compile dependency towards `xbean-asm9-shaded`. The `org.ow2.asm` stuff is a transient dependency of `xbean-asm9-shaded`. So shouldn't we exclude  `xbean-asm9-shaded` here as it comes from `openwebbeans` anyway?

container/openejb-core/pom.xml
```diff
@@ -562,11 +562,31 @@
...
+        <exclusion>
+          <groupId>org.ow2.asm</groupId>
+          <artifactId>asm-commons</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.xbean</groupId>
       <artifactId>xbean-finder-shaded</artifactId>
+      <exclusions>
```
rzo1, line 580: see above

arquillian/arquillian-tomee-common/pom.xml
rmannibucau: ow2 deps should be excluded in root pom in dependency management of xbean-asm9-shaded dependency declaration to ensure the whole tree gets it for free.

cocorossello: Ok, let me change it tomorrow. Thanks for the hint, didn't really now where to exclude or if all exclusions were needed

cocorossello: Changed it, only that exclusion is enough

rzo1 (approved):

mawiesne (approved):

PR #779, opened 2021-04-03
Fix some examples using BOM

cocorossello:


cocorossello:
Still not working:
examples/polling-parent/polling-web/pom.xml
examples/applicationcomposer-jaxws-cdi/pom.xml

examples/polling-parent/pom.xml
```diff
@@ -79,29 +79,17 @@
...
-        <artifactId>openejb-core</artifactId>
-        <version>${tomee.version}</version>
-        <scope>provided</scope>
-      </dependency>
-      <dependency>
-        <groupId>org.apache.tomee</groupId>
-        <artifactId>openejb-cxf-rs</artifactId>
-        <version>${tomee.version}</version>
+        <groupId>org.apache.tomee.bom</groupId>
+        <artifactId>tomee-plus-api</artifactId>
```
rzo1, line 91: Should probably be `tomee-plus` instead (otherwise duplicate). Would also go for scope `test` here (as `test` would be finde for standard code but we might need to override it in submodules)

examples/polling-parent/polling-web/pom.xml
```diff
@@ -35,21 +35,14 @@
...
     <dependency>
-      <groupId>org.apache.tomee</groupId>
-      <artifactId>openejb-core</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.tomee</groupId>
-      <artifactId>openejb-cxf-rs</artifactId>
+      <groupId>org.apache.tomee.bom</groupId>
+      <artifactId>tomee-plus</artifactId>
+      <version>8.0.7-SNAPSHOT</version>
```
rzo1, line 40: No version needed. Tree gets it for free from the parent dep mgmt section.

examples/polling-parent/polling-domain/pom.xml
```diff
@@ -29,8 +29,9 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.apache.tomee</groupId>
-      <artifactId>javaee-api</artifactId>
+      <groupId>org.apache.tomee.bom</groupId>
+      <artifactId>tomee-plus-api</artifactId>
+      <version>8.0.7-SNAPSHOT</version>
```
rzo1, line 34: No version needed. Tree gets it for free from the parent dep mgmt section.

examples/polling-parent/polling-client/pom.xml
```diff
@@ -30,27 +30,9 @@
...
-    <dependency>
-      <groupId>org.apache.xbean</groupId>
-      <artifactId>xbean-finder-shaded</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.xbean</groupId>
-      <artifactId>xbean-reflect</artifactId>
+      <groupId>org.apache.tomee.bom</groupId>
+      <artifactId>tomee-plus</artifactId>
+      <version>8.0.7-SNAPSHOT</version>
```
rzo1, line 35: No version needed. Tree gets it for free from the parent dep mgmt section.
If you go for `test` in the parent, we need to set it to `provided` here as we have some dependencies.

examples/polling-parent/polling-client/pom.xml
```diff
@@ -62,8 +44,9 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
-      <groupId>org.apache.tomee</groupId>
-      <artifactId>javaee-api</artifactId>
+      <groupId>org.apache.tomee.bom</groupId>
+      <artifactId>tomee-plus-api</artifactId>
+      <version>8.0.7-SNAPSHOT</version>
```
rzo1, line 49: No version needed. Tree gets it for free from the parent dep mgmt section.

examples/polling-parent/polling-core/pom.xml
```diff
@@ -29,20 +29,18 @@
...
     <dependency>
-      <groupId>org.apache.tomee</groupId>
-      <artifactId>javaee-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.tomee</groupId>
-      <artifactId>openejb-api</artifactId>
+      <groupId>org.apache.tomee.bom</groupId>
+      <artifactId>tomee-plus-api</artifactId>
+      <version>8.0.7-SNAPSHOT</version>
```
rzo1, line 34: No version needed. Tree gets it for free from the parent dep mgmt section.

examples/polling-parent/polling-core/pom.xml
```diff
@@ -29,20 +29,18 @@
...
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.apache.tomee</groupId>
-      <artifactId>openejb-core</artifactId>
+      <groupId>org.apache.tomee.bom</groupId>
+      <artifactId>tomee-plus</artifactId>
+      <version>8.0.7-SNAPSHOT</version>
```
rzo1, line 43: No version needed. Tree gets it for free from the parent dep mgmt section.
We have some dependencies towards `openjpa` here, so we would need `provided` here (only we have `test` in the parent pom)

examples/polling-parent/polling-web/pom.xml
rzo1: In addition: `provided` if `test` in parent pom.

examples/polling-parent/pom.xml
```diff
@@ -79,29 +79,17 @@
...
-        <version>[8.0,)</version>
-        <scope>provided</scope>
-      </dependency>
-      <dependency>
-        <groupId>org.apache.tomee</groupId>
-        <artifactId>openejb-api</artifactId>
-        <version>${tomee.version}</version>
+        <groupId>org.apache.tomee.bom</groupId>
+        <artifactId>tomee-plus-api</artifactId>
+        <version>8.0.7-SNAPSHOT</version>
```
rzo1, line 84: Maybe use `tomee.version` instead of the hard string as it is already defined in the `properties` section. We also can remove `cxf.version` and `xbean.version` from the `<properties` section of the parent pom. 

In addition, we can remove

```
 <dependency>
        <groupId>org.apache.cxf</groupId>
        <artifactId>cxf-core</artifactId>
        <version>${cxf.version}</version>
      </dependency>
      <dependency>
        <artifactId>xbean-finder-shaded</artifactId>
        <version>${xbean.version}</version>
      </dependency>
      <dependency>
        <artifactId>xbean-reflect</artifactId>
        <version>${xbean.version}</version>
      </dependency>
```

as it comes with the related BOM.

examples/applicationcomposer-jaxws-cdi/pom.xml
```diff
@@ -64,10 +64,10 @@
...
   <dependencies>
     <dependency>
-      <groupId>org.apache.tomee</groupId>
-      <artifactId>javaee-api</artifactId>
-      <version>[8.0,)</version>
-      <scope>provided</scope>
+      <groupId>org.apache.tomee.bom</groupId>
+      <artifactId>tomee-plus-api</artifactId>
+      <version>8.0.7-SNAPSHOT</version>
+      <scope>test</scope>
```
rzo1, line 70: Shouldn't it be `provided` ?

rzo1: Otherwise:

It seems, that the tests are failing as the `tomee-plus` boms brings in `tomee-security`, which expects some servlet registrations in `TomEESecurityServletAuthenticationMechanismMapper` and subsequently fails with a `NullPointerException`. 

So either

- (a) The test is flawed and requires some additional parameters / registrations
- (b) `TomEESecurityServletAuthenticationMechanismMapper` requires a ` != null` check on `servletRegistrations` in `init(...)`
- (c) Exclude `tomee-security` in the examples dependencies
- (d) Something else (disable security in the example, other things, ...) ....

Maybe @dblevins has some thoughts here?

rzo1 (commented):
Thanks @cocorossello - I would go for ` <scope>provided</scope>` for every `-api` module.
My comments focus on the **non** working examples.

examples/applicationcomposer-jaxws-cdi/pom.xml
dblevins: My recommendation would be to for both C and B.  Specifically, do the exclude (B) so we can get the build working, then execute C and remove the exclude.

cocorossello: Excluded, thanks, I didn't know whether servletRegistrations should be not null or (more likely) servletRegistrations should not be called

rzo1: @cocorossello The `api` module needs to be in scope `provided` in this example. Besides that: Works fine for me locally :) (and all other examples are passing the tests too... so overall lgtm. thx!)

rzo1 (approved):

dblevins:
Not sure why Jenkins didn't build this PR after it was updated, but merging to see if we can fix the build.

Thank you @cocorossello

rzo1:
I opened a PR to test the integration a few days ago, which worked. @cocorossello uses a shared repository for the PRs - might be the reason why Jenkins does not pick up these PRs - would be imho an odd behaviour of the Jenkins Plugin.

PR #795, opened 2021-05-11
Added back test step for gihub action

jeanouii:


rzo1:
Think we can close this one. This has diverged from current GH actions setup.

PR #798, opened 2021-05-21
Fix example adoc typo

Dexmaster:
For comparison check last tests block https://github.com/Dexmaster/tomee/blob/master/examples/rest-cdi/README.adoc vs https://github.com/apache/tomee/blob/master/examples/rest-cdi/README.adoc


rzo1:
Nice find. Thx.

PR #801, opened 2021-08-26
Restore missing Inline quoted text for tomee-maven-plugin.adoc

mkienenb:
Inline quoted text seems to have been lost in the conversion from MD to adoc.   For unknown reasons it's missing on the web page but visible in view-source for https://tomee.apache.org/tomee-maven-plugin.html, which makes me wonder if this is really the source file or if the web page hasn't been redeployed since the conversion.


rzo1 (approved):
Thanks for the fix.

PR #826, opened 2022-03-25
TOMEE-3872 - Update Hibernate Integration to 5.6.7

rzo1:
Updates Hibernate to 5.6.7.

Currently having a full build on Jenkins - so let's what will happen.


rzo1:
**Full** build is **green** now! Think, we can go ahead with this one.

rzo1:
Rebased and squased into one commit.

PR #827, opened 2022-03-25
TOMEE-3876 - Fixed BOM generation under windows

sultan:
now group and artifact ids have dots instead of slashes. now ok under windows like its was already ok on linux and mac


rzo1:
@sultan You need to create this PR against `tomee-8.x` as we switched our master branch to JakartaEE 9.1 just yesterday ;) Nevertheless, thank you for creating it in the first place!

rzo1:
@sultan I changed the base branch, so we are good now. Let's wait for GitHub actions and then I will merge it (and pick it for master as well)

rzo1 (approved):

PR #828, opened 2022-03-26
TOMEE-3871 - TomEE Plume is missing BatchEE / JCS Cache

sultan:
batchee and jcs added in Plume to reflect Plus flavor

removed xml api from from Plume to reflect Plus flavor
remains some jar refs in jar scan exclusions (xerces, xml-apis)

### DIFF
--- a/boms/tomee-microprofile/pom.xml
+++ b/boms/tomee-microprofile/pom.xml
@@ -2125,17 +2125,6 @@
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>xml-apis</groupId>
-      <artifactId>xml-apis</artifactId>
-      <version>1.4.01</version>
-      <exclusions>
-        <exclusion>
-          <artifactId>*</artifactId>
-          <groupId>*</groupId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-    <dependency>
       <groupId>xml-resolver</groupId>
       <artifactId>xml-resolver</artifactId>
       <version>1.2</version>
--- a/boms/tomee-plume/pom.xml
+++ b/boms/tomee-plume/pom.xml
@@ -332,6 +332,17 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>org.apache.batchee</groupId>
+      <artifactId>batchee-jbatch</artifactId>
+      <version>1.0.1</version>
+      <exclusions>
+        <exclusion>
+          <artifactId>*</artifactId>
+          <groupId>*</groupId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
       <groupId>org.apache.bval</groupId>
       <artifactId>bval-jsr</artifactId>
       <version>2.0.5</version>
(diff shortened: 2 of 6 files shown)


rmannibucau:
AFAIK plume is a webprofile based distro so batchee and jcs should belong to a new "plume-all" distro probably?

rzo1:
> AFAIK plume is a webprofile based distro so batchee and jcs should belong to a new "plume-all" distro probably?

This originates from this [list thread](https://lists.apache.org/thread/o6tqmsgmc975ddr028ohtw2r8npmzg0m) and @dblevins comment there, that: "Those distributions are supposed to be the same minus the JPA and JSF providers.". There is also a [blog post](https://www.tomitribe.com/blog/tomee-webprofile-vs-tomee-microprofile-vs-tomee-vs-tomee-plume/) about the differences but I am too new to judge the original thoughts regarding "plume" vs "plus".

sultan:
> AFAIK plume is a webprofile based distro

if this is correct then Plume should not contains jars for MicroProfile nor JMS ActiveMQ

image

image

rmannibucau:
Yep, think they shouldnt but as Im not a big user of that I dont care much of what it becomes while webpro and plus stays aligned on their goal ;)

sultan:
webpro and plus have a clear focus to me as well, even though "Plus" branding could be changed to better tell its "Full EE" 🤔 or i can just describe that in the flavors page.

You could drop Plume but you would lose the benefits from its "ready to try out" capabilities.
i am thinking about the AutoConfiguration for Jakarta Persistence (JPA) which facilitates learning Tomee.

rzo1:
I think it is good to have this discussion to clarify what the intended distributions should look like in 2022 considering the history of Plus / Plume and the incosistent current status of the distros / website. So thanks for the additional thoughts @rmannibucau and @sultan

sultan:
does this need a mailing list vote from committers ?

dblevins:
Here's the list message on the name and motivation for Plume:

 - https://lists.apache.org/thread/tkpfjlzv70y8kzz5f905qj9vs95b34rw

"a version of TomEE Plus with Mojarra and EclipseLink. Tentatively called it "TomEE Plume" which stands for TomEE PLUs, Mojarra, Eclipselink (PLU.M.E)."

sultan:
nice to read !
1) i learned that a setup installer for TomEE "like eclipse oomph" (cherry pick modules) is not possible, as it breaks TCK rules. unless we want a parallel non TCK certified TomEE (meh)
2) i thought like @rmannibucau at first with "per functionalities" LIB folders (like others EE servers) but this would make TomEE harder to configure / extend. (where do user put their jars !?)
3) there could be "per functionalities" uber or shaded jars to ease remove and replace implementations by others. but i wonder if i did read that current project policy is to avoid shading/uber.
4) there could be a common POM between plus and plume to help fill the gaps/ease maintenance.
5) there could be an external tool to modify a dezipped certified TomEE to change its implementation jars.

the only one i think i can currently help on is n°4 (common plus-plume pom)

rzo1 (approved):

rmannibucau:
> there could be an external tool to modify a dezipped certified TomEE to change its implementation jars.

If you look tomee-maven-plugin it enables to do any distro from plain tomcat to advanced customized tomee and you can use it to build a tomee plume from tomee webprofile or any distro ;). Installer is legacy since years :)

rzo1:
@sultan Can you check your rebase with `tomee-8.x`. It does not look sane to me (at least judging from the changeset now).

sultan:
i think i was able to clean up my mess by reverting to my previous draft. commits seems legit now. can someone launch checks ?

PR #830, opened 2022-03-27
TOMEE-3647

cesarhernandezgt:


rzo1 (approved):
@cesarhernandezgt lgtm. We should probably bring this to master / 9.x as well.

cesarhernandezgt:
> @cesarhernandezgt lgtm. We should probably bring this to master / 9.x as well.

Hi @rzo1 , 
I tried today to apply the patch to master (9.x) but maybe we can stabilize the first master branch since I haven't been able to have a successful build even by skipping all the tests. wdyt?

rzo1:
We can do either. Master will fail with most of the examples anyway due to outdated dependencies without jakarta modifier /classifier.

I agree, that it is a bit of a pain to ship something in unknown waters 😀 my comment aims for only one thing: we should 
not forget it.

cesarhernandezgt:
> my comment aims for only one thing: we should
> not forget it.
Totally agree, I created https://issues.apache.org/jira/browse/TOMEE-3883 aiming at master (9) so we not forget.

PR #855, opened 2022-04-17
TOMEE-3846 Add Comparison page for TomEE v9 master

sultan:
this is the "per-version" comparison page for v9


sultan:
ready for review and open to feedbacks

sultan:
commits squashed

PR #866, opened 2022-04-20
TOMEE-3846 Add Comparison page for TomEE v7.1

sultan:
this is the "per-version" comparison page for TomEE v7.1

i have several questions/issues:
* i lack some JSR specs number, i need to dig them, working on it, help appreciated
* this comparison page would be a quasi duplicate between 7.0 and 7.1, it is ok for now.
  * there are 2 TomEE minors (v7.0,v7.1) for 1 Java EE (v7) its ok if it stays like that
  * it may be a pain later if there is again 2 TomEE for 1 Jakarta (say TomEE v10.0 and TomEE v10.1 for same Jakarta 10.0)
  * could this happen again ?
* i would appreciate feedbacks on this suggested comparison page
  * is red cross a bad idea to tell EE7 does not include JSON-B ?
  * do i better remove the lines with red crosses (specs not part of EE7) ?

instead of making a comparison page for every TomEE minors (7.0, 7.1, 8, 9), 
could i make a comparison page for every JavaEE/JakartaEE minors (7, 8, 9.1, 10.0) ?
(and put them on the main website instead)

i'm not sure whats the best option.

here is what i'am working on : 
- tomee.apache.org/
  - comparison.html
  - tomee-7.1/docs/comparison.html (more risk ok duplicate)
  - tomee-8.0/docs/comparison.html
  - tomee-9.0/docs/comparison.html

here is an alternative : 
- tomee.apache.org/
  - comparison.html
  - comparison-javaee-7.0.html (no risk ok duplicate)
  - comparison-jakartaee-8.0.html
  - comparison-jakartaee-9.1.html
  - comparison-jakartaee-10.0.html


sultan:
Jakarta EE 8 has RPC, does Jakarta EE 9 and 11 include RPC 1.1.4 ?

dblevins:
@sultan on the spec version questions, some of those are a bit tricky.  For example ActiveMQ only implements JMS 1.1, however there is code in TomEE to implement some of the JMS 2.0 features.  I don't know if there are any gaps.  We're not running the JMS TCK so it's hard to say.

Going to merge these comparison PRs and if we need to adjust the versions we list, we can do that in separate PRs

PR #875, opened 2022-04-29
TOMEE-3931 fixed cucumber-jvm example and added deps/cucumber-openejb-shade module.

cesarhernandezgt:


rzo1 (commented):
I am wondering about the re-generated BOM files. Adding a new shade dep shouldn't alter them? Perhaps an older (?) local state? Or did we alter the TomEE Jakarta API snapshot recently?

The rest of the PR lgtm.

cesarhernandezgt:
> I am wondering about the re-generated BOM files. Adding a new shade dep shouldn't alter them? Perhaps an older (?) local state? Or did we alter the TomEE Jakarta API snapshot recently?

I wondered the same, I pull the latest master and after building without a test but with -Pno-examples, boms are updated. I opened a separate PR to double check that this was not related to adding a shade dep: https://github.com/apache/tomee/pull/876

If 876 is fine, we can merge and I'll rebase pr 875, or we can just go ahead with 875 and I would drop 876

rzo1 (approved):

rzo1:
I did merge #876, which led to https://github.com/apache/tomee/pull/877 by our automatic action, which should be the source of truth. If I am building master locally, I do not get any changed BOMs 🤔 Can you try with `-U` to update local SNAPSHOT artifacts?

cesarhernandezgt:
Thank you for the review and help @rzo1 . You were right, updating snapshot by running locally mvn with -U removed the poms update from the PR. I rebase the PR now with only the changes for TomEE-3931

rzo1 (approved):

rzo1:
Hi @cesarhernandezgt ,

the PR looks good to me - I think we can proceed with it.

While testing it  locally, I noticed, that the README did not reflect the changes. Therefore, I went a head and updated it with the new versions / dependencies.

cesarhernandezgt:
Nice, thank you!. Merging now...

rmannibucau:
Side note: can be worth doing it at cucumber directly instead of managing a fork ;)

rzo1:
@rmannibucau You are right - I was thinking the same. In the long run, it would be better to have it in cucumber. Therefore, I have opened an issue https://github.com/cucumber/cucumber-jvm/issues/2548, so we can discuss the possible ways of providing a Jakarta artifact at cucumber directly.

PR #876, opened 2022-04-29
Updated boms from current master

cesarhernandezgt:


cesarhernandezgt:
I pull current master and after executing 
`clean install -DskipTests -Pno-examples`
The resulting boms are generated.

PR #883, opened 2022-05-10
Tomee 3824

tichovz:
I create a new method into ObjectFactory and a new test method, where unmarshalling the web-fragment.xml is successful.
Sorry I made a code format in JeeTest.java. 
Unfortunately I do not know how to test ObjectFactory.createWebFragment method.

### DIFF
--- a/container/openejb-jee/src/main/java/org/apache/openejb/jee/ObjectFactory.java
+++ b/container/openejb-jee/src/main/java/org/apache/openejb/jee/ObjectFactory.java
@@ -55,6 +55,7 @@ public class ObjectFactory {
     private final static QName _JavaWsdlMapping_QNAME = new QName("http://java.sun.com/xml/ns/j2ee", "java-wsdl-mapping");
     private final static QName _Webservices_QNAME = new QName("http://java.sun.com/xml/ns/javaee", "webservices");
     private final static QName _FacesConfig_QNAME = new QName("http://java.sun.com/xml/ns/javaee", "faces-config");
+    private final static QName _WebFragment_QNAME = new QName("http://java.sun.com/xml/ns/javaee", "web-fragment");
 
     /**
      * Create a new ObjectFactory that can be used to create new instances of schema derived classes for package: org.apache.openejb.jee2
@@ -165,5 +166,13 @@ public class ObjectFactory {
     @XmlElementDecl(namespace = "http://java.sun.com/xml/ns/javaee", name = "faces-config")
     public JAXBElement<FacesConfig> createFacesConfig(final FacesConfig value) {
         return new JAXBElement<FacesConfig>(_FacesConfig_QNAME, FacesConfig.class, null, value);
+    }
+    
+     /**
+     * Create an instance of {@link JAXBElement }{@code <}{@link WebFragment }{@code >}}
+     */
+    @XmlElementDecl(namespace = "http://java.sun.com/xml/ns/javaee", name = "web-fragment")
+    public JAXBElement<WebFragment> createWebFragment(final WebFragment value) {
+        return new JAXBElement<WebFragment>(_WebFragment_QNAME, WebFragment.class, null, value);
     }
 }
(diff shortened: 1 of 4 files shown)


cesarhernandezgt:
Hi @tichovz , thank you for the PR.
The PR is against apache:develop (last updated back in 2015); perhaps you may want to open this against the current Master branch?

rzo1:
I think you are right @cesarhernandezgt if I am following the discussion on the list correctly. @tichovz Would you mind chaning the target to `master` `

tichovz:
Hi!

I changed the target

container/openejb-jee/src/test/java/org/apache/openejb/jee/JeeTest.java
```diff
@@ -18,23 +18,22 @@
 
 package org.apache.openejb.jee;
 
+import jakarta.xml.bind.*;
```
rzo1, line 21: Can we restore the original import order? It is basically a style change

container/openejb-jee/src/test/resources/web-fragment-example.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rzo1, line 1: This file misses a license header - subsequently, the license header checks in CI will fail. 

You can ake a look at the other descriptor files and just copy & paste the ASF license header into this file.

container/openejb-jee/src/main/java/org/apache/openejb/jee/ObjectFactory.java
```diff
@@ -166,4 +167,12 @@ public JAXBElement<Webservices> createConnector(final Webservices value) {
     public JAXBElement<FacesConfig> createFacesConfig(final FacesConfig value) {
         return new JAXBElement<FacesConfig>(_FacesConfig_QNAME, FacesConfig.class, null, value);
     }
+    
+     /**
+     * Create an instance of {@link JAXBElement }{@code <}{@link WebFragment }{@code >}}
+     */
+    @XmlElementDecl(namespace = "http://java.sun.com/xml/ns/javaee", name = "web-fragment")
```
rzo1, line 174: We might need to add `web-fragment` part to `ObjectFactory$JAXB` in `openejb-jee-accessors` to fully support it?

rzo1:
Thanks for the PR @tichovz - I left some comments.

container/openejb-jee/src/test/java/org/apache/openejb/jee/JeeTest.java
tichovz: I restored the original order, but I have not commited yet.

container/openejb-jee/src/test/resources/web-fragment-example.xml
tichovz: I pasted the license header, but I have not commited yet.

container/openejb-jee/src/main/java/org/apache/openejb/jee/ObjectFactory.java
tichovz: I added the web-fragment part to ObjectFactory$JAXB, but I did not solve the issue, because the org.apache.openejb.config.ReadDescriptors.mergeWebFragments calls JaxbJavaee.unmarshall which uses the NoSourceFilter.
It does not translate the namespace "https://jakarta.ee/xml/ns/jakartaee" to "http://java.sun.com/xml/ns/javaee".
So what would I have to do?
1. change  ReadDescriptors.mergeWebFragments to call JaxbJavaee.unmarshallJavaee which changes the namespace
2. change the NoSourceFilter to change the jakarta namespace as well not only the "http://xmlns.jcp.org/xml/ns/javaee"
3. anything else 

I think I might remove the web-fragment part from ObjectFactory$JAXB

rzo1: Thanks for the comment. I think, that the best way would be to follow up on the list. @dblevins might have some thoughts / insights here.

rzo1 (approved):

tichovz:
Hi!

Can I do something to make the pull request successful?

Best :
Zoltán

rzo1:
@tichovz Thanks!

PR #891, opened 2022-06-25
Tomee 3955

tichovz:


examples/arquillian-jpa/pom.xml
```diff
@@ -105,6 +106,33 @@ language governing permissions and limitations under the License. -->
...
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13.2</version>
```
rzo1, line 133: It looks like the tests were migrated towards JUnit 5 - do we still need JUnit4 here?

examples/arquillian-jpa/pom.xml
```diff
@@ -28,10 +28,22 @@ language governing permissions and limitations under the License. -->
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.5.1</version>
```
rzo1, line 29: Can we  use `3.10.1` ?

examples/arquillian-jpa/pom.xml
```diff
@@ -28,10 +28,22 @@ language governing permissions and limitations under the License. -->
...
       </plugin>
+      <plugin>
+       <groupId>org.apache.maven.plugins</groupId>
+       <artifactId>maven-surefire-plugin</artifactId>
+       <version>2.22.2</version>
+       <dependencies>
+                <dependency>
+                    <groupId>org.junit.jupiter</groupId>
+                    <artifactId>junit-jupiter-engine</artifactId>
+                    <version>5.7.0</version>
```
rzo1, line 43: Any reason for `5.7.0` vs `5.8.2` ?

examples/arquillian-jpa/pom.xml
```diff
@@ -28,10 +28,22 @@ language governing permissions and limitations under the License. -->
...
-          <source>1.8</source>
-          <target>1.8</target>
+          <source>11</source>
+          <target>11</target>
         </configuration>
       </plugin>
+      <plugin>
+       <groupId>org.apache.maven.plugins</groupId>
+       <artifactId>maven-surefire-plugin</artifactId>
+       <version>2.22.2</version>
```
rzo1, line 38: Can we use `3.0.0-M5` or higher here?

examples/arquillian-jpa/src/test/resources/arquillian.xml
```diff
@@ -25,7 +25,7 @@
     <property name="defaultDataSource">openejb:Resource/demoDataSource</property>
   </extension>
 
-  <container qualifier="tomee" default="true">
+  <container qualifier="tomee-embedded" default="true">
```
rzo1, line 28: Looks like it changed from embedded to remote?`If so, we might need to update the `README` as well?

rzo1 (commented):
@tichovz Thanks for the PR.

examples/arquillian-jpa/pom.xml
tichovz: Yes, we still need it, because without it a "java.lang.ClassNotFoundException: org.junit.runners.BlockJUnit4ClassRunner"
occurs during the loading of tomee-embedded container

examples/arquillian-jpa/pom.xml
tichovz: Yes, we can, I ugraded the dependency

examples/arquillian-jpa/pom.xml
tichovz: There is no particular reason the use of 5.7.0 so I upgraded it to 5.8.2

examples/arquillian-jpa/pom.xml
tichovz: Yes, we can, so I upgraded it to 3.0.0-M7

examples/arquillian-jpa/src/test/resources/arquillian.xml
tichovz: I may see it wrong, but I think I’ve switched from remote to embedded. If I use the remote container, the injection of the persistence context and the @EJB annotation are not working (the entitity manager is null and the transactionalCaller is also null

rzo1: You are right. My bad ;)

tichovz:
Hi!

Do I need to update the README? If so, could anyone help me with where that
file is and what should I write in it?

Thanks: Zoltán

rzo1:
@tichovz The README files are here https://github.com/apache/tomee/tree/master/examples/arquillian-jpa (basically same level as the POM)

PR #900, opened 2022-07-06
Update exception message for MDB without an interface

bdemers:
I don't actually know what a `MDB` is, after a quick google i'm guessing it a `MessageDriven` bean".  Either way, the error message should spell it out.

> **NOTE:** I updated this through GitHub's web view, if CI fails, and this change is _valid_ I'll try to fix the patch.

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/cdi/CdiEjbBean.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/cdi/CdiEjbBean.java
@@ -250,7 +250,7 @@ public class CdiEjbBean<T> extends BaseEjbBean<T> implements InterceptedMarker, 
             } else if (remote != null) {
                 instance = (T) remote.create();
             } else { // shouldn't be called for an MDB
-                throw new IllegalStateException("no interface to proxy for ejb " + beanContext.getEjbName() + ", is this is a @MessageDriven bean, maybe you shouldn't use a scope?");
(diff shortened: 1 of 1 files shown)


container/openejb-core/src/main/java/org/apache/openejb/cdi/CdiEjbBean.java
```diff
@@ -250,7 +250,7 @@ protected T createEjb(final CreationalContext<T> creationalContext) {
             } else if (remote != null) {
                 instance = (T) remote.create();
             } else { // shouldn't be called for an MDB
-                throw new IllegalStateException("no interface to proxy for ejb " + beanContext.getEjbName() + ", is this is a MDB maybe you shouldn't use a scope?");
+                throw new IllegalStateException("no interface to proxy for ejb " + beanContext.getEjbName() + ", is this is a @MessageDriven bean, maybe you shouldn't use a scope?");
```
bdemers, line 253: In my case, it's not a message-driven bean, but I like the suggestion, maybe the wording should be something like:

```suggestion
                throw new IllegalStateException("No interface to proxy for ejb " + beanContext.getEjbName() + ", if this is a @MessageDriven bean, try not using a scope?");
```

rzo1: Thanks @bdemers - you are right. MDB means `@MessageDriven` bean here. I think that the proposed change is valid and clarifies the error message.

PR #907, opened 2022-07-31

sultan:
Tomcat does not include JSTL nor the javax.security.enterprise.* packages.
The documentation should reflect that, and should have these specs into WebProfile.

### DIFF
--- a/docs/comparison.adoc
+++ b/docs/comparison.adoc
@@ -88,7 +88,7 @@ https://hibernate.org/validator/[Hibernate Validator^] 7.0.x *(in TomEE 9.x and 
 |Jakarta Enterprise Beans (EJB)|https://openejb.apache.org/[Apache OpenEJB^]
 |Jakarta Persistence (JPA)|
 https://openjpa.apache.org/[Apache OpenJPA^] 3.2.x with classifier (in all TomEE flavors) +
-https://www.eclipse.org/eclipselink/[EclipseLink^] 3.1.x *(in TomEE Plume only)*
+https://www.eclipse.org/eclipselink/[EclipseLink^] 3.x *(in TomEE Plume only)*
 |Jakarta Transactions (JTA)|Apache{nbsp}Geronimo{nbsp}Transaction{nbsp}Manager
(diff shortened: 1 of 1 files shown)


rzo1 (approved):

rzo1:
Thanks @sultan

dblevins:
Sorry all for not seeing this PR sooner.  Here's information on Tomcat's JSTL and JASPIC implementations:

 - https://tomcat.apache.org/taglibs.html
 - https://tomcat.apache.org/tomcat-9.0-doc/config/jaspic.html

We should probably revert this change.

rzo1:
> We should probably revert this change.

@dblevins Missed that. I will do the reverts.

sultan:
the tomcat website indeed provides implementation for JSTL, but its not included in the tomcat server release itself.

rzo1:
I just did the reverts as suggested by @dblevins 

Perhaps, it would be good to move this discussion towards the dev@ list, so we get consensus of what we want to display in our docs?

dblevins:
@sultan you're right -- looks like they're not shipping it and we have a shaded version of that jar as it isn't even converted to the jakarta namespace.  Also looks like I got JASPIC and Jakarta Security mixed up.

@rzo1 sorry for the busywork

rzo1:
@dblevins no problem.

How do we want to proceed? Looks like the changes are valid (but maybe not clear enough?) and we have a lot of confusion around it. So it can only improve :)

- (A) We recreate the PRs + the PR in site generator, ask for review on the list and move on an after some time
- (B) We recreate the PRs + the PR in site generator and move on 
- (C) Revert the revert, i.e. cherry pick the commit from the squash merge and move on

In terms of a clear git history, I would prefer (A) or (B) but we can also do (C), if this is less effort.

sultan:
i think i wrongly moved debugging support out of tomcat into webprofile spec. so i made some minor changes that will require me to make the PRs again. i can take advantage of the new PRs to add some text to about tomcat not shipped with JSTL nor security api packages

PR #909, opened 2022-07-31

sultan:
Tomcat does not include JSTL nor the javax.security.enterprise.* packages.
The documentation should reflect that, and should have these specs into WebProfile.


rzo1 (approved):

rzo1:
Thanks @sultan

PR #913, opened 2022-08-27
example code for TomEE beginner

air-h-128k-il:
Hi, i am new to TomEE.

i think TomEE is a great project.
I tried some sample code, but the coding style for testing in `examples` was too modern to understand for me.

Therefore, i have converted your async-servlet project into a simple web application.

async-servlet2-tomee

Wouldn't it be nice to have some projects for beginners?

### DIFF
--- a/examples-for-beginner/async-servlet2/.gitignore
+++ b/examples-for-beginner/async-servlet2/.gitignore
@@ -1,0 +1,3 @@
+target/
+.DS_Store
+nb-configuration.xml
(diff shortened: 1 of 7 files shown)


jeanouii:
Hi,

Thanks for the PR.
Small note: please move the example to examples/ directory so they are all grouped together. Also our tooling for the website is made such as this directory is scanned. 

Also please find a proper name instead of async-servlet2.

Finally, sorry for the dumb question, but I can't see what is the difference aside from the /calc added in the URL and the test case that you removed.

Can you elaborate on what was the problem you are trying to solve?

air-h-128k-il:
Hi,

Thank you for your kind response.

> I can't see what is the difference aside from the /calc added in the URL and the test case that you removed

That's right.
I didn't mean to add any original flavor.

I'm sorry if I offended you, but very few people around me knew about TomEE.
For those people we just wrote the examples out in a .war file the classic way.

We will withdraw a pull request if it doesn't fit the concept.



2022年9月14日(水) 21:55 Jean-Louis Monteiro ***@***.***>:

PR #914, opened 2022-08-28

sultan:


sultan:
Here is an extract of the current visual result : 
image
image

sultan:
Thanks @jeanouii ! 

https://github.com/apache/tomee-site-generator/pull/43 is still pending (same improvements) and https://github.com/apache/tomee-site-generator/pull/48 also pending

PR #924, opened 2022-09-14
[TOMEE-4035] - MicroProfile 5.0

sultan:
MicroProfile 5.0 & SmallRye Fault Tolerance 6.0.0 instead of RC

Fixed Copy paste error for Metrics

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -229,8 +229,7 @@
 
     <microprofile.fault-tolerance.version>4.0</microprofile.fault-tolerance.version>
     <microprofile.fault-tolerance.tck.version>4.0</microprofile.fault-tolerance.tck.version>
-    <microprofile.fault-tolerance.impl.version>6.0.0-RC4</microprofile.fault-tolerance.impl.version>
-
+    <microprofile.fault-tolerance.impl.version>6.0.0</microprofile.fault-tolerance.impl.version>
 
     <microprofile.health.version>4.0</microprofile.health.version>
     <microprofile.health.tck.version>4.0</microprofile.health.tck.version>
--- a/pom.xml
+++ b/pom.xml
@@ -225,7 +225,7 @@
 
     <microprofile.jwt.version>2.0</microprofile.jwt.version>
     <microprofile.jwt.tck.version>2.0</microprofile.jwt.tck.version>
-    <microprofile.jwt.impl.version>${project.version}</microprofile.jwt.impl.version>
+    <microprofile.jwt.impl.version>4.0.0</microprofile.jwt.impl.version>
 
     <microprofile.fault-tolerance.version>4.0</microprofile.fault-tolerance.version>
     <microprofile.fault-tolerance.tck.version>4.0</microprofile.fault-tolerance.tck.version>
--- a/pom.xml
+++ b/pom.xml
@@ -225,7 +225,7 @@
 
     <microprofile.jwt.version>2.0</microprofile.jwt.version>
     <microprofile.jwt.tck.version>2.0</microprofile.jwt.tck.version>
-    <microprofile.jwt.impl.version>4.0.0</microprofile.jwt.impl.version>
+    <microprofile.jwt.impl.version>${project.version}</microprofile.jwt.impl.version>
 
     <microprofile.fault-tolerance.version>4.0</microprofile.fault-tolerance.version>
     <microprofile.fault-tolerance.tck.version>4.0</microprofile.fault-tolerance.tck.version>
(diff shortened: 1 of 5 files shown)


tomee/tomee-microprofile/mp-common/pom.xml
```diff
@@ -27,6 +27,128 @@
...
+      </dependency>
+      <dependency>
+        <groupId>io.smallrye</groupId>
+        <artifactId>smallrye-health</artifactId>
+        <version>${microprofile.health.impl.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.smallrye</groupId>
+        <artifactId>smallrye-jwt</artifactId>
+        <version>4.0.0</version>
```
rzo1, line 113: any reason for not having a property here?

sultan: it seems that TomEE does not use SmallRye Impl for JWT.
however this could be useful for outside users that need a BOM for MP 5.0 with all impls.

i'll rework a property in the main POM.

pom.xml
```diff
@@ -225,12 +225,11 @@
 
     <microprofile.jwt.version>2.0</microprofile.jwt.version>
     <microprofile.jwt.tck.version>2.0</microprofile.jwt.tck.version>
-    <microprofile.jwt.impl.version>${project.version}</microprofile.jwt.impl.version>
+    <microprofile.jwt.impl.version>4.0.0</microprofile.jwt.impl.version>
```
jeanouii, line 228: I don't think this is what we want.
We have our own JWT implementation with more features than SmallRye so we use our own for this one.

tomee/tomee-microprofile/mp-common/pom.xml
```diff
@@ -27,6 +27,128 @@
...
+        <version>${microprofile.health.impl.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.smallrye</groupId>
+        <artifactId>smallrye-health</artifactId>
+        <version>${microprofile.health.impl.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.smallrye</groupId>
+        <artifactId>smallrye-jwt</artifactId>
```
jeanouii, line 112: See previous comment. We don't need this.

tomee/tomee-microprofile/mp-common/pom.xml
```diff
@@ -84,7 +206,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>mp-jwt</artifactId>
-      <version>${microprofile.jwt.impl.version}</version>
+      <version>${project.version}</version>
```
jeanouii, line 209: See previous comment. If you don't change the property, it has ${project.version} already.

tomee/tomee-microprofile/mp-common/pom.xml
```diff
@@ -138,7 +260,7 @@
     <dependency>
       <groupId>org.eclipse.microprofile.metrics</groupId>
       <artifactId>microprofile-metrics-api</artifactId>
-      <version>${microprofile.health.version}</version>
+      <version>${microprofile.metrics.version}</version>
```
jeanouii, line 263: Good catch but if you have dependency management, shouldn't we take the version out?

jeanouii (changes requested):

jeanouii (changes requested):
See comments. Nothing crazy.

Thanks for the PR

jeanouii (changes requested):
See comments. Nothing crazy.

Thanks for the PR

pom.xml
sultan: rolledback this change.

tomee/tomee-microprofile/mp-common/pom.xml
sultan: removed this.

tomee/tomee-microprofile/mp-common/pom.xml
sultan: rolledback this change.

tomee/tomee-microprofile/mp-common/pom.xml
sultan: took out all the managed versions

jeanouii (approved):
Thank for the quick turn around. I'm fine with the PR. Waiting for the build to give us a confirmation and then we can merge

rzo1 (approved):

rzo1:
Building `main` as well as this branch works on my Windows 10 vm. I do not get the issue described in [TOMEE-4045](https://issues.apache.org/jira/browse/TOMEE-4045). Let's see, if Jenkins picks up the PR as PR builds tend to be unreliable on Jenkins in the last months. Otherwise, we might be able to "just merge" this in as the GitHub actions are already green, so no compile issues.

jeanouii:
I'm fin in merging and fixing after if needed

rzo1:
Thanks @sultan

PR #927, opened 2022-09-27
[TOMEE-4035] Jakarta 9.1.0 and MicroProfile 5.0

sultan:
[TOMEE-4052] - SAAJ Impl 2.0.1 => see https://issues.apache.org/jira/browse/TOMEE-4052
Krazo 2.0.2 => same for both EE 9.1 and 10.0 !
CXF 3.5.3
MP Config 3.0.2
MP FT 4.0.1

Log4J 2.19.0


sultan:
This is a proposal, open to discussion. I am available on demand to create one Jira ticket per artifact.

PR #928, opened 2022-09-27
[TOMEE-4053] TomeE 9.x Dependency properties cleanup

sultan:
Versions not to be upgraded:
* cxf = 3.5.0 and not 3.5.4

Changed versions:
* jackson = 2.14.0-rc2

Removed properties:

    <pax-url.version>1.3.5</pax-url.version>
    <wagon.version>2.2</wagon.version>
    <plexus.version>1.5.5</plexus.version>
    <plexus-utils.version>3.0.10</plexus-utils.version>
    <org.apache.axis2.version>1.4.1</org.apache.axis2.version>
    <scannotation.version>1.0.2</scannotation.version>
    <geronimo-osgi.version>1.1</geronimo-osgi.version>

### DIFF
--- a/arquillian/arquillian-openejb-embedded/pom.xml
+++ b/arquillian/arquillian-openejb-embedded/pom.xml
@@ -170,7 +170,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
-        <version>${surefire.version}</version> <!-- 2.21 snapshot has a bug with classloader due to j9 work -->
+        <version>${version.plugin.surefire}</version> <!-- 2.21 snapshot has a bug with classloader due to j9 work -->
       </plugin>
     </plugins>
   </build>
--- a/arquillian/arquillian-tck/pom.xml
+++ b/arquillian/arquillian-tck/pom.xml
@@ -97,7 +97,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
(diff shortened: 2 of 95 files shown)


jeanouii:
Hi,

Yes by nature the big PRs are difficult to review. That being said, I appreciate the effort to bring clarity and consistency.
I'm ok to merge it and see if it breaks and possibly fix afterwards.

Would you please start an email on the mailing list, to make sure everyone can voice his opinion?

Thanks

dblevins:
I think it's fantastic to get the version strings normalized to `version.foo`.  We started with `foo.version` and gradually changed to `version.foo` as that kind of thing became more favored by Maven in general, but we never went back and fixed the old properties so we have a mix.  Great to see that addressed.

My preference would be to keep it short and go with straight `version.foo` versus adding prefixes before the `foo` part such as `tomee`, `api` or the groupId.  I'll never remember which prefix rule goes where and when, so I'll constantly be looking it up and also further adding to inconsistency.

Intellij does have completion for Maven properties.  Being able to type `version.` and then get a flat list of all the names would be fantastic.

sultan:
> Looks much better to me!

using shorter names instead of groupid everywhere looks a bit easier to read, i kept the longer names branch on my computer if we ever need them.

> @sultan Looks much better to me! Can we yank the "eclipse" before all the "microprofile" references? On the MicroProfile side of the fence we generally dislike "Eclipse MicroProfile" and always ask people to change it.

done

sultan:
ready for CI build/tests, might consider merging this one before:
* PR #940

sultan:
#928 (this PR) and #936 superseded by other PRs

PR #932, opened 2022-09-30
Improve compatibility matrix to include DeltaSpike as an Apache alternative to Spring Data JPA

sultan:
image

### DIFF
--- a/docs/comparison.adoc
+++ b/docs/comparison.adoc
@@ -107,17 +107,19 @@ https://smallrye.io/[SmallRye MicroProfile^] {y} *(ok with TomEE 9.x and later)*
 
 In bold : Implementations that differ between flavors or between versions
 
-== [[Compatibility]] Compatibility with other implementations
+== [[Compatibility]] Implementations compatibility and alternatives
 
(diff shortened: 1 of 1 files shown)


dblevins:
Looking at this, we should maybe create another page that's dedicated to using popular libraries or implementations in TomEE where we can be more helpful and a little more accurate.

Hibernate definitely is an implementation of Jakarta Persistence, whereas Spring Web MVC does not implement Jakarta MVC, and neither Delta Spike nor Spring Data JPA implement Jakarta Data as that spec really hasn't started yet and can't be used by anyone.

I think the goal is great, but a table doesn't really communicate the various states and this page is maybe getting too large.

sultan:
yep definitely not implementations but alternatives to the lack of them.

dblevins:
Just thought of another idea.  What if we create a page called `implementations.adoc` and move all the implementation and alternative implementation information there?  (thinking out loud)

That would thin comparison page a bit.  Of course we could link the implementation page from the bottom of the comparison page.

Thoughts?

PR #942, opened 2022-10-09
TOMEE-4057 - Remove CXF patch sources with no TomEE-related modifications

rzo1:
# What does this PR do?

During the CXF upgrade to 3.4.8  and porting the method signature changes of CXF-8478 and CXF-8688 to our  CXF patch sources, I took some time to check the **diffs** between our patch sources and the original CXF sources of 3.4.8.

Our patch sources were indeed a bit outdated. Some of them didn't have any modifications from our TCK work in 2021 at all (aside from some formatting, language style, etc.).

Therefore, I went ahead and removed all CXF classes from our patch source, which didn't contain "real" changes from our side by doing a diff session. 

This will give us the fixes in those classes, which we didn't incorporated through simple POM version updates since May 2021.
Goal is to make it easier to upgrade CXF, i.e. to check the diffs of the classes we did modify on purpose.

Here is the full CI build on that one: https://ci-builds.apache.org/job/Tomee/job/TOMEE-4057/2/


rmannibucau:
Hi

Cxf just published a 4.0.0-SNAPSHOT targetting jakarta, can be worth a try to drop all the fork at once?

rzo1:
Hi @rmannibucau 

We are indeed discussing CXF 4 for TomEE 9.x / 10.x on the list: https://lists.apache.org/thread/o21s1gk07m3khyx5nmyxbx7xforxknro

For 8.x, it could be worth to clean up the Jakarta transform / shading things. We don't need them in 8.x  as we now have a 9.x targeting the namespace and aren't torn to use jakarta relocation anymore. So perhaps time to cleanup :-)

The remaining patch sources are related to some TCK tests / provider ordering afaik. I would need to look that up, check the mailing list or git log for information and track / check, if these changes are now contained in CXF itself, so we can drop in 8.x 

I am +1 in dropping the fork sooner or later because it confuses automation / scanners (+ perhaps users).

PR #944, opened 2022-10-20
Javax to jakarta on deploy

jeanouii:
Small quick and dirty hacking to test on the fly conversion of javax apps using Arquillian or Tomcat webapp-javaee/ directory. For completeness, we need also Maven and probably some other entry points, but it's mainly a show case to show it's possible.

The counter part part is slow startup.


jeanouii:
This seems to be too dangerous and with too many restrictions to move on for now

PR #947, opened 2022-10-23
TOMEE-4088 TomEE 9.x + hsqldb 2.7.1 instead of workaround for CVE-2022-41853

sultan:
removed workaround upon dependency update hsqldb 2.7.1


rzo1:
Thanks @sultan

Current tomee-8.x includes the updates and the removal of the workaround. So closing this draft.

PR #950, opened 2022-10-26
Adapt to new Tomcat threadNameBindings, use Map Interface instead of …

Florian-Rau:
…implementation


Florian-Rau:
Fix for https://issues.apache.org/jira/browse/TOMEE-4102

rzo1:
@Florian-Rau Thanks for detailed reporting and providing a patch!

rzo1 (approved):

PR #953, opened 2022-10-28
[TOMEE-4053] Dependency properties cleanup, step 1 and 2 on branch 9.x

sultan:
this branch has the properties re-ordered/regrouped by categories, with name changes, we can test this on ci build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,6 @@
     <!-- allows release plugin to be overridden -->
     <arguments />
     <tomee.version>${project.version}</tomee.version>
-    <surefire.version>3.0.0-M7</surefire.version>
 
     <!-- for the default name of the module. Needs to be overridden -->
     <tomee.build.name>${project.groupId}.${project.artifactId}</tomee.build.name>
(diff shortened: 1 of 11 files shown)


rzo1 (approved):

sultan:
to be merged after 
* PR #957

sultan:
rebased and ready

sultan:
added a step 2 with a few name changes that passed OK on the previous version branch

pom.xml
```diff
@@ -1341,7 +1327,7 @@
       <dependency>
         <groupId>org.apache.geronimo.mail</groupId>
         <artifactId>geronimo-mail_2.1_provider</artifactId>
-        <version>${geronimo-mail_2.1_provider.version}</version>
+        <version>${version.geronimo-javamail_1.6_mail}</version>
```
rzo1, line 1330: This version string doesn't look right? Might explain my confusion regarding one of the dropped properties.

pom.xml
```diff
@@ -102,28 +101,16 @@
     <version.jakartaee-api>9.1-M2</version.jakartaee-api>
     <version.tomee-patch-plugin>0.9</version.tomee-patch-plugin>
 
-    <!--
-      JavaMail is both API and IMPL so we don't have them in the jakartaee-api uber jar.
-      We decided to add them here in the project so we can patch/update the 2 libraries without having
```
rzo1, line 107: Guess we should keep these comments

pom.xml
```diff
@@ -102,28 +101,16 @@
...
-      to update and release the jakartaee-api.jar. Different lifecycle.
-    -->
-    <geronimo-mail_2.1_spec.version>1.0.0-M1</geronimo-mail_2.1_spec.version>
-    <geronimo-mail_2.1_provider.version>1.0.0</geronimo-mail_2.1_provider.version>
```
rzo1, line 111: Any reason for dropping one of these versions? Guess we need to check, what is needed / required.

pom.xml
sultan: The name was incorrect, my mistake. now corrected

pom.xml
sultan: relevant question, they are not dropped but moved, i redid the PR again and fixed the props names

pom.xml
rzo1: Moving away from Geronimo API + Impls need to be discussed on the list. 

I wouldn't change that in this PR. We should keep it as ist and do the cleanup/renaming. If we have consensus on the list, we might switch away.

sultan: to be clear my intent is to keep Geronimo Impl. and swtich only the API.
but i get your point and will discuss that on the mailing list.

rzo1: There is also an ongoing discussion regarding Geronimo components on their list. There might be changes in the near future (attic, etc), but that discussion hasn't ended yet. So we might need to discuss / change things in the future.

rzo1 (approved):

PR #954, opened 2022-10-28
[TOMEE-4053] Dependency properties cleanup, step 2 on branch 8.x

sultan:
this one has some properties renamed, only when they are used within the same root pom.

these name changes may be tested against ci tests if needed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,8 +107,8 @@
 
     <!-- Plugins and Build -->
     <maven.version>3.0.5</maven.version>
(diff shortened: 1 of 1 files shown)


rzo1:
CI for this one is running. Let's see what will happen next ;)

sultan:
my ide returned no other files impacted, unless there are string concatenation, which i believe would be bad.

expect step3 and step4 to impact microprofile :)

rzo1 (approved):

PR #955, opened 2022-10-30
[TOMEE-4053] Dependency properties cleanup, step 3 on branch 8.x

sultan:
this one is about tck tests and microprofile api & impl.

### DIFF
--- a/tck/microprofile-tck/pom.xml
+++ b/tck/microprofile-tck/pom.xml
@@ -63,4 +63,72 @@
       </modules>
     </profile>
   </profiles>
+
+  <dependencyManagement>
+    <dependencies>
+      <!-- Micro Profile TCK -->
+      <dependency>
+        <groupId>org.eclipse.microprofile.config</groupId>
+        <artifactId>microprofile-config-tck</artifactId>
+        <version>${version.microprofile.config}</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.eclipse.microprofile.fault-tolerance</groupId>
+        <artifactId>microprofile-fault-tolerance-tck</artifactId>
+        <version>${version.microprofile.fault-tolerance}</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.eclipse.microprofile.health</groupId>
+        <artifactId>microprofile-health-tck</artifactId>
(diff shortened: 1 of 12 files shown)


sultan:
there may be room in another step to factorize more the microprofile api & tck dependencies and exclusions

tck/microprofile-tck/pom.xml
```diff
@@ -63,4 +63,134 @@
       </modules>
     </profile>
   </profiles>
+
+  <dependencyManagement>
```
rzo1, line 67: Any reason for adding them under `dependencyManagement` in the `microprofile-tck`  again? The only difference is the `scope = test`, which is decleared on each artifact in the related sub-modules?

tomee/tomee-microprofile/mp-common/pom.xml
```diff
@@ -27,7 +27,61 @@
   <artifactId>mp-common</artifactId>
   <name>TomEE :: TomEE :: MicroProfile Common</name>
 
+  <dependencyManagement>
```
rzo1, line 30: same comment as above

rzo1:
CI Build is running for this one

sultan:
yes i can do better on this part, maybe factorize the exclusions or remove the doubling of deps managements

rzo1:
Any updates?

sultan:
sorry for the lack of update, i'll have time to rework the PR this weekend or after the daily job. it should not take me much time, but i'll switch the draft to ready only when ok.

sultan:
all the versions are locked in a parent pom when desirable. there are dependency management sections for MP in three pom:
* root pom.xml for MP API
* mp-common pom.xml for MP Geronimo Impl
* microprofile-tck for MP TCK tests

tck/microprofile-tck/pom.xml
sultan: i believe this is not a duplicate, there are artifacts for API and more artifacts for TCK tests.

tomee/tomee-microprofile/mp-common/pom.xml
sultan: i believe this is not a duplicate, there are artifacts for API and more artifacts for Impl.

rzo1 (approved):

PR #956, opened 2022-11-02
TOMEE-4104 - Fixed bug in MySQL example code.

stoxford:
[TOMEE-4104](https://issues.apache.org/jira/browse/TOMEE-4104). Added missing &lt;/Resource&gt; tag to MySQL example code.


rzo1:
Thanks @stoxford

PR #959, opened 2022-11-08
TOMEE-4065 Fix to allow @LoginToContinue to be used directly in the HttpAuthenticationMechanism

jeanouii:
instead of requiring the interface to be implemented. Thanks Stefan Kalscheuer for reporting and for the fix

### DIFF
--- a/itests/pom.xml
+++ b/itests/pom.xml
@@ -49,6 +49,7 @@
     <module>jaxrs</module>
     <module>itest-common</module>
     <module>itest-util</module>
+    <module>tomee-security-itests</module>
   </modules>
 
   <build>
--- a/itests/tomee-security-itests/pom.xml
+++ b/itests/tomee-security-itests/pom.xml
@@ -1,0 +1,111 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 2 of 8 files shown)


stklcode (approved):
Looks good to me. Thanks very much for the fix!

I tested my target application with both current (non-annotated) and original (annotated) implementation against the snapshot build (Microprofile with Java 17, but the result should be the very same for other flavors). Both working as expected.


Think I personally would have added a regression test case for `@FormAuthMechanismDefinition` with nested annotation, although that's already partially covered by other test suites (like https://github.com/stklcode/tomee/commit/a06dc9e600b522195874fcb515da360cd77a48bf).

jeanouii:
I agree that it does not hurt. Can you push your changes in this branch so when I merge, your changes are included instead of me stealing your test?

stklcode:
I do not have write permissions on this repository, so I'm afraid I cannot do this.

What I can do is (re-)create this PR from my fork or you can cherry-pick the referenced commit.

jeanouii:
You should be able to create a PR against this branch so I can merge it

rzo1:
@jeanouii Do we want to backport this change to 8.x, too ?

jeanouii:
Yes we can. It should work the same way

rzo1:
> Yes we can. It should work the same way

Ok. I will create a Jira for it, so we can pick it up.

PR #963, opened 2022-11-09
[TOMEE-4053] Dependency properties cleanup, step 4 on branch 8.x

sultan:
this one is renaming Jakarta EE version properties

### DIFF
--- a/server/openejb-cxf/pom.xml
+++ b/server/openejb-cxf/pom.xml
@@ -194,7 +194,7 @@
     <dependency>
       <groupId>org.apache.cxf</groupId>
       <artifactId>cxf-rt-frontend-jaxws</artifactId>
-      <version>${cxf.version}</version>
+      <version>${version.cxf}</version>
       <exclusions>
         <exclusion>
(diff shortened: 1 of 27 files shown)


server/openejb-cxf/pom.xml
```diff
@@ -264,7 +264,7 @@
     <dependency>
       <groupId>com.sun.xml.messaging.saaj</groupId>
       <artifactId>saaj-impl</artifactId>
-      <version>${saaj-impl.version}</version>
+      <version>${version.impl.saaj}</version>
```
rzo1, line 267: `version.saaj.impl` or `version.saaj-impl` ?

sultan: The intent was to have all future ambiguous impl in a "version.impl" prefix like version.impl.activation and version.impl.mail to disambiguate with api version with prefixes without impl. There should be ambiguity only with those three. It can be changed to something else entirely if needed.

rzo1 (approved):

PR #976, opened 2022-11-20
[TOMEE-4115] Tomcat update to 9.0.69

sultan:
TomEE 8.x dependency update to tomcat 9.0.69 (EE 8)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -200,7 +200,7 @@
     <version.geronimo-javamail_1.6_spec>1.0.1</version.geronimo-javamail_1.6_spec>
 
     <!-- Jakarta EE Impl. -->
-    <tomcat.version>9.0.68</tomcat.version>
+    <tomcat.version>9.0.69</tomcat.version>
     <!-- com.sun -->
     <version.impl.saaj>1.5.1</version.impl.saaj>
     <!-- org.apache -->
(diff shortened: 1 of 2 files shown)


rzo1:
Changelog: https://tomcat.apache.org/tomcat-9.0-doc/changelog.html

```java
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.2:compile (default-compile) on project tomee-catalina: Compilation failure
Error:  /home/runner/work/tomee/tomee/tomee/tomee-catalina/src/main/java/org/apache/tomee/catalina/LazyStopStandardRoot.java:[38,8] org.apache.tomee.catalina.LazyStopStandardRoot is not abstract and does not override abstract method getArchiveIndexStrategyEnum() in org.apache.catalina.WebResourceRoot
```

It seems this update requires some changes due to

>  Add a configuration option to allow bloom filters used to index JAR files to be retained for the lifetime of the web application. Prior to this addition, the indexes were always flushed by the periodic calls to WebResourceRoot.gc(). As part of this addition, configuration of archive indexing moves from Context to WebResourceRoot. Based on a patch provided by Rahul Jaisimha. (markt)

in https://github.com/apache/tomee/blob/tomee-8.x/tomee/tomee-catalina/src/main/java/org/apache/tomee/catalina/LazyStopStandardRoot.java

It might also need a check, if this change needs some additional changes in that module to fully incorporate it.

sultan:
note to self:
* look for TomEE docs whether its allowed or not to use AOP for delegates.

PR #984, opened 2022-12-09
TOMEE-4014 alternative approach that does not touch ServerInfo.proper…

jgallimore:
…ties in catalina.jar, and works around the reflection issue with the final modifier


rzo1 (approved):
lgtm

PR #990, opened 2022-12-18
Implement tomee.mp.jwt.allow.no-exp property over mp.jwt.tomee.allow.…

tichovz:
…no-exp

### DIFF
--- a/mp-jwt/src/main/java/org/apache/tomee/microprofile/jwt/config/JWTAuthConfigurationProperties.java
+++ b/mp-jwt/src/main/java/org/apache/tomee/microprofile/jwt/config/JWTAuthConfigurationProperties.java
@@ -38,6 +38,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 
 import static org.eclipse.microprofile.jwt.config.Names.AUDIENCES;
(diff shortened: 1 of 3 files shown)


tichovz:
Should I change the property name in the JWT TCK?

https://github.com/apache/tomee/blob/main/tck/microprofile-tck/jwt/src/test/java/org/apache/tomee/microprofile/tck/jwt/validation/ExpClaimAllowMissingExpValidationTest.java#L89

rzo1:
Do we have an issue for this one @tichovz ?

tichovz:
Hi!

I have no issues for this, but David asked if could take a look at it.

rzo1:
Thanks. That was the missing link, so maybe you can just rebase and we do a CI build to see, if it breaks anything?

tichovz:
Thanks, Would I make a rebase with the main branch or the 9.x branch?

rzo1:
Let's test against main and if it looks nice, we should backport to 9.1.x too

tichovz:
I made a rebase to main and it seems ok.

mp-jwt/src/main/java/org/apache/tomee/microprofile/jwt/config/JWTAuthConfigurationProperties.java
```diff
@@ -117,6 +119,15 @@ private JWTAuthConfiguration createJWTAuthConfiguration() {
                 config.getOptionalValue("mp.jwt.decrypt.key.algorithm", String.class).orElse(null),
                 config.getOptionalValue("mp.jwt.verify.publickey.algorithm", String.class).orElse(null));
     }
+    
+    private Boolean queryAllowExp(){
```
rmannibucau, line 123:
```
    private Boolean queryAllowExp(){
        return config.getOptionalValue("tomee.mp.jwt.allow.no-exp", Boolean.class)
                .or(() -> config.getOptionalValue("mp.jwt.tomee.allow.no-exp", Boolean.class)
                        .map(value -> {
                            CONFIGURATION.warning("mp.jwt.tomee.allow.no-exp property is deprecated, use tomee.mp.jwt.allow.no-exp propert instead.");
                            return value;
                        }))
                .orElse(false);
    }
```

to avoid to read both entries all the time (`Config` can be slow depending the `ConfigSource`) and to avoid the `AtomicBoolean` which is not needed?

tichovz: Thanks for the improvement.

rzo1: @tichovz Can you include the feedback provided by Romain? :)

PR #991, opened 2022-12-18
[TOMEE-4128] Jakarta MVC API and Krazo (Impl.) provided in TomEE 8.1.x 'plume' and 'plus' server flavors

sultan:
This PR modifies the dependencies for plume and plus, adding the following dependencies:
* Jakarta MVC API
* Krazo CXF

this allows an application to avoid depending on the implementation choices (CXF)
now an app can depend on the API instead.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -226,6 +226,7 @@
     <!-- Other API and Impl. not in Jakarta EE -->
     <version.ehcache>2.10.6</version.ehcache>
     <version.geronimo-jcache_1.0_spec>1.0-alpha-1</version.geronimo-jcache_1.0_spec>
+    <version.mvc>1.1.0</version.mvc>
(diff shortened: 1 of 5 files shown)


dblevins:
If we were to add more specs to the TomEE 8.x branch, we'd need to release that as 8.1.x like we did when we created TomEE 7.1.x due to adding the MicroProfile specs.  We should probably discuss on the dev list if we want to do that.

sultan:
> If we were to add more specs to the TomEE 8.x branch, we'd need to release that as 8.1.x like we did when we created TomEE 7.1.x due to adding the MicroProfile specs. We should probably discuss on the dev list if we want to do that.

that would be cleaner yes 🙂
making this draft until i can add the tck

rzo1:
8.x is eol now.

PR #993, opened 2022-12-19
TOMEE-4014 - Re-add "tomee.keep-server-info"

rzo1:
as the title says


rzo1:
@jgallimore Judging from the answer in the ticket, it seems sufficient to re-add the check on "tomee.keep-server-info".

rzo1:
@jgallimore  Switched from `System.getProperties()` to the original code using the `SystemInstance#getOptions()`, so it should be compatible to the way it previously worked.

PR #997, opened 2023-01-04
[TOMEE-4053] Dependency properties cleanup, step 3 : Micro Profile properties on branch main/10.x

sultan:
[TOMEE-4053] Dependency properties cleanup, step 3 : Micro Profile properties on branch main/10.x

Edit: now without changes to the versions numbers

### DIFF
--- a/deps/cxf-rt-rs-mp-client-shade/pom.xml
+++ b/deps/cxf-rt-rs-mp-client-shade/pom.xml
@@ -65,7 +65,6 @@
     <dependency>
       <groupId>org.eclipse.microprofile.rest.client</groupId>
       <artifactId>microprofile-rest-client-api</artifactId>
-      <version>${microprofile.rest-client.version}</version>
     </dependency>
 
   </dependencies>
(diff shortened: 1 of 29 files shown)


sultan:
@jeanouii here is the work i did and still doing for both properties cleanup and update MP to 6.0

sultan:
i was a bit agressive on the update. should i revert my changes on fault tolerance fallback (fallback to 1.0 ?)

rzo1:
Does th MP 6.0 TCK pass? If not, we should do the property cleanup and afterwards update to a (passing) MP6

sultan:
> Does th MP 6.0 TCK pass? If not, we should do the property cleanup and afterwards update to a (passing) MP6

split the PR into separate PRs

this one is for the properties names changes. without versions number chnages

rzo1:
I think, that the target should be https://github.com/apache/tomee/pull/1005

sultan:
@rzo1 rebased and merged

jeanouii:
Thanks @sultan

PR #1012, opened 2023-01-31
[TOMEE-4053] Dependency properties cleanup, on branch for tomcat 10 and cxf 4

sultan:


sultan:
in case you need the actions to run. i can simply alter the conditions with *

PR #1016, opened 2023-01-31
[TOMEE-4182] GitHub actions: 8.x is not main. keeping only 8.x files

sultan:
8.x is not main. removing main file from non-main branches


rzo1:
thx

PR #1017, opened 2023-01-31
[TOMEE-4182] post release 9.0.0

sultan:
this should get the build to succeed but in wonder if its ok to merge it


rzo1:
I merged the post release fix from Cesar, so it's green on 9.x - if still needed, we should merge this too

PR #1021, opened 2023-02-07
[TOMEE-4061] Derby 10.15.2.0, branch 10.x

sultan:
This is a tentative to update Derby to 10.15.x, upon other dep updates.

Derby 10.15.x is modular, it might break CI, please merge wisely.

TomEE 9.x dependencies update: 
* Derby 10.15.2.0


sultan:
if ever the build would break because if this, we might be able to fix adding another dependency.
id'like to see if it works without; but unable to build, looks like missing deps in patch and or tomee-javaee.

    <dependency>
      <groupId>org.apache.derby</groupId>
      <artifactId>derbytools</artifactId>
      <version>${version.derby}</version>
      <scope>test</scope>
    </dependency>

only if needed

jeanouii:
The branch is currently actively being worked on. I'll start doing milestone releases from Jakarta EE API and others when possible so we can get people to build and the CI at some point

Long story short, I'll merge it without the extra dependency and we can fix later if that's ok with you

sultan:
thank you very much

PR #1023, opened 2023-02-13
[TOMEE-4053] Dependency properties cleanup, branch 10.x

sultan:
this one is about microprofile.

next PR will be upgrading to MP 6.0 (all except SmallRye Metrics)


rzo1:
Actually, most changes were already integrated. Closing.

PR #1025, opened 2023-02-27
TOMEE-4189 - java.lang.ClassNotFoundException: org.apache.openejb.loader.SystemInstance

rzo1:
# What does this PR do?

The classpath for the `version` command of  `./catalina[.sh|.bat] ` does only contain

```bash
 -classpath "$CATALINA_HOME/lib/catalina.jar"  org.apache.catalina.util.ServerInfo
```

Thus, our reflection code introduced with https://github.com/apache/tomee/commit/7ce518e29111ba10cd6bcf5a8fc613fc912b84a8 and https://github.com/apache/tomee/commit/14a034cc9528247d2c30c052412e730d64dea094 will fail with a `ClassNotFoundException` as `openejb-core` isn't available in the given classpath.

However, this is the only occurence within  `./catalina[.sh|.bat] ` , so it should be ok to adjust the classpath for the given command only.


cesarhernandezgt (approved):
Thank you @rzo1 . 
I tested the PR and tested `./bin/catalina.sh version` and now no java.lang.ClassNotFoundException were found.

PR #1026, opened 2023-03-12
TOMEE-4190 RunWithApplicationComposer should support inheritance

G-Ork:
- Added Inherited annotation.
- Added unittests to proove fix.


rzo1 (approved):
Thanks for your contribution, @G-Ork !

PR #1059, opened 2023-06-27
TOMEE-4226 - Check for modern PostgreSQL driver, if JDBCUrl is not specified in @DataSourceDefinition

rzo1:
# What does this PR do?

- Check for modern PostgreSQL driver, if JDBCUrl is not specified in @DataSourceDefinition
- Fixes TOMEE-4226: https://issues.apache.org/jira/browse/TOMEE-4226
- Introduces `DataSourceDefinitionUrlBuild` to "plugin" custom `@Event` listener to adjust datasource definition, if needed.

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/config/AnnotationDeployer.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/config/AnnotationDeployer.java
@@ -23,6 +23,7 @@ import org.apache.openejb.api.LocalClient;
 import org.apache.openejb.api.Proxy;
 import org.apache.openejb.api.RemoteClient;
 import org.apache.openejb.cdi.CdiBeanInfo;
+import org.apache.openejb.config.event.DataSourceDefinitionUrlBuild;
 import org.apache.openejb.config.rules.CheckClasses;
 import org.apache.openejb.core.EmptyResourcesClassLoader;
 import org.apache.openejb.core.ParentClassLoaderFinder;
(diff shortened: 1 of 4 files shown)


rzo1:
@cesarhernandezgt @jeanouii This can also happen for **MariaDB** driver, so I am wondering, if we should include that here as well (might be more common than **MySQL** these days). wdyt?

rmannibucau:
if it helps, my two cents would be it can be worth adding a DataSourceDefinitionUrlBuild openejb event since these guessed URL are rarely prod friendly urls so firing the event with the data we have (from the annot) and allowing to set an URL can be a way to enable users to customize the way the url is built (plugin like) then supporting by default all world patterns become less important since there are several cases we just have to do a choice so 50% it can be wrong ;).

rzo1:
If I understand the idea correctly, that would mean, that we create something like:

```bash

package org.apache.openejb.config.event;

import org.apache.openejb.jee.DataSource;
import org.apache.openejb.observer.Event;

@Event
public class DataSourceDefinitionUrlBuild {

    private final DataSource ds;

    public DataSourceDefinitionUrlBuild(DataSource ds) {
        this.ds = ds;
    }

    //getters
}

```
and fire it as soon as we build a `org.apache.openejb.jee.DataSource` from `@DataSourceDefinition` in `AnnotationDeployer#buildDataSourceDefinition(...)` ?

Someone could then do something like:

```bash
    public void adjustDataSourceDefinition(@Observes final DataSourceDefinitionUrlBuild event) {
      //do something (for example set a custom jdbc url).
    }
````

and register that class via `org.apache.openejb.extension` in `META-INF`, right?

rzo1:
I went ahead and did a test impl for the `@Event` `DataSourceDefinitionUrlBuild` (at least from my understanding of your idea :-) )

rmannibucau (approved):

PR #1073, opened 2023-10-26
Fix #TOMEE-4265: EarClassLoaderTest and EarWebAppFirstClassLoaderTest 

exabrial:
Fix #TOMEE-4265: EarClassLoaderTest and EarWebAppFirstClassLoaderTest test failures on 8.0.16-SNAPSHOT. joda-time dependency was upgraded from 2.5 and the corresponding tests were not upgraded to 2.10.10


rzo1:
I think the difference is in purpose. It should use the older version first (as it is bundled within the webapp) and than fallback to the newer version.

The test typically fails due to flakyness om CI.

exabrial:
Got it, I'll close this, thanks!

PR #1081, opened 2023-11-17
add an integration test for optional beans

struberg:
if a class creates a NoClassDefFoundError or ClassNotFoundException then it's simply not an eligible bean class.

All works fine in TomEE 10!


itests/openejb-itests-optional-classes/3rdpartydep/src/main/java/org/apache/tomee/itests/optional/thirdparty/ThirdPartyApi.java
```diff
@@ -0,0 +1,10 @@
+package org.apache.tomee.itests.optional.thirdparty;
```
rzo1, line 1: @struberg Missed header.

PR #1086, opened 2023-12-03
TOMEE-4284 - Implement tomee.mp.jwt.allow.no-exp property over mp.jwt.tomee.allow.no-exp

tichovz:
This is instead of pull request #990: TOMEE-4284 - Implement tomee.mp.jwt.allow.no-exp property over mp.jwt.tomee.allow.no-exp
https://github.com/apache/tomee/pull/990


tichovz:
I included the feedback provided by Romain. But this code does not give deprecation warning when both properties present
in the config. If this is not a problem, I will remove this requirement from the itests, because there are two tests that expect a warning to be given in this case as well. 

David asked for this: "any use of mp.jwt.tomee.allow.no-exp should get a warning log message"

PR #1093, opened 2024-01-16
Improve failed startup follow-up exceptions and back to OWB snapshot

tandraschko:
OWB contains new 2 fixes, one is necessary for JSF applications as some ELs are not resolveable anymore

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -218,7 +218,7 @@
     <version.johnzon>2.0.0</version.johnzon>
     <version.myfaces>4.0.1</version.myfaces>
     <version.openjpa>4.0.0-SNAPSHOT</version.openjpa>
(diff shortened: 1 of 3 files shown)


rzo1:
@tandraschko From my understanding this should fix https://github.com/apache/tomee/blob/main/arquillian/arquillian-tomee-moviefun-example/src/test/java/org/superbiz/moviefun/MoviesHtmlUnitTest.java#L79 - maybe we can re-add that test?

tandraschko:
@rzo1 yep
i can make a new PR later, or if you have 1 min, would be great can you do it

rzo1:
Yeah, I will merge this and re-enable the test.

PR #1114, opened 2024-03-25
TOMEE-4305 | TOMEE-4306 - Tomcat Patches

rzo1:
as the title says. Ports patches from Tomcat

@tandraschko Can you have a look?


rzo1:
Triggered full build: https://ci-builds.apache.org/job/Tomee/view/tomee-9.x/job/pull-request-9.x-manual/31/

jungm:
The diff of `WsSession.java` in https://github.com/apache/tomee/pull/1114/commits/5420188ed2263e2dec491d29d049f21155d28252#diff-7d73ff8a9a8066d1da614c448a86924480535b514e50a137600675316464354d is a lot larger than the one in https://github.com/apache/tomcat/commit/0052b374684b613b0c849899b325ebe334ac6501#diff-a25788433a677ecd33c43b5b90c8d24c15bd4ee20cc76fd0948807c7750f7cbf (therefore contains a lot of changes that haven't been made in the upstream tomcat commit)

Possibly just copy & pasted entire file instead of just applying the patch?
Everything else looks good to me

rzo1:
Yes. I did copy more than the fix as this class was subject of heavy changes in 10.1.x - I think, it was ok to do so, but open to opinions.

jungm:
Absolutely fine for me as well, it just stood out to me

gerdogdu:
@rzo1 This type of patching is really hard to follow. Needs to find a better way such as creating a profile for TCK which includes the patches otherwise, no patch needed but not TCK certified.

rzo1:
@gerdogdu We only port CVE related things atm. It is sad, that Tomcat stopped maintaining 10.0.27, so we just cannot backport every bug fix, change, feature, which was added since 10.0.27 (resources, time, etc.). There are plenty of differences in (older) patch files between newer Tomcat releases. Feel free to start a discussion on the mailing list, if we want to drop TCK compliance and just upgrade to 10.1.x on the TomEE 9.1.x line. That way, we could just stop porting patches for an ancient piece of unmaintained software.

rzo1:
I might find some time in the upcoming weeks to investigate, if we can upgrade without hurting the 9.1 tck ;-)

gerdogdu:
> @gerdogdu We only port CVE related things atm. It is sad, that Tomcat stopped maintaining 10.0.27, so we just cannot backport every bug fix, change, feature, which was added since 10.0.27 (resources, time, etc.). There are plenty of differences in (older) patch files between newer Tomcat releases. Feel free to start a discussion on the mailing list, if we want to drop TCK compliance and just upgrade to 10.1.x on the TomEE 9.1.x line. That way, we could just stop porting patches for an ancient piece of unmaintained software.

I think we should focus on the using latest Tomcat 10 release instead of only gets CVE patches, it is so absurd. How do you ensure that getting only CVE patches not break the old Tomcat APIs? 

EE 9 is a intermediate release with API change. I think we should drop TCK compatibility for 9 and look TCK compatibility for EE 10.

rzo1:
Yup, I agree in principal. But, this discussion needs to be started on the dev list. If you like, you can start the discussion there. Meanwhile, I started some tests with 9.x with 10.1.x to see  TCK Impact :-)

rmannibucau:
If anyone interested there is a thread started today on tomcat list to see if we can get a 10.0.
Not very confident it will end up on something but having a certifiable version is probably still key for tomee so 10.1 is not yet an option - will be hopefully.
That said we can certainly request an exception to certify v9 with servlet 6 instead of 5 - guess it is possible with some warnings?

rzo1:
Most likely to just ignore signature tests 😄

PR #1120, opened 2024-04-13
TOMEE-4318 | TOMEE-4317 - ActiveMQ 6 + Java 17

rzo1:
- https://issues.apache.org/jira/browse/TOMEE-4318
- https://issues.apache.org/jira/browse/TOMEE-4317

-> https://ci-builds.apache.org/job/Tomee/job/pull-request-manual/84/


container/openejb-loader/src/main/java/org/apache/openejb/loader/Embedder.java
```diff
@@ -105,6 +105,7 @@ private Class forcefulLoad(final ClassPath classPath, final ClassLoader classLoa
             }
             classPath.addJarsToPath(libsDir);
         } catch (final Exception e2) {
+            e2.printStackTrace();
```
jungm, line 108: Probably unwanted?

tomee/tomee-webaccess/pom.xml
```diff
@@ -42,6 +42,19 @@
         <activeByDefault>true</activeByDefault>
       </activation>
     </profile>
+    <profile>
+      <id>jdk15+</id>
+      <activation>
+        <jdk>[15,)</jdk>
```
jungm, line 48: imo no reason to hide this dependency inside of a profile if the compile level is raised to 17 (also applies to java11plus profile above)

jungm (commented):
Just some small details i noticed 🙂

container/openejb-loader/src/main/java/org/apache/openejb/loader/Embedder.java
rzo1: yes

PR #1130, opened 2024-05-15
TOMEE-4332 Patch SpringWebUtils for class loader issue

jgallimore:


jgallimore:
Closing and replacing with #1165

PR #1131, opened 2024-05-15
TOMEE-4332 Patch SpringWebUtils for class loader issue

jgallimore:


tandraschko:
we are on the newest CXF SNAPSHOT, doesnt it make sense to fix it there?

jgallimore:
Closed and replaced with #1166

PR #1167, opened 2024-05-25
TOMEE-4342 - ApplicationComposer in PER_JVM should be able to inject @Resource into test

rzo1:


mawiesne (approved):
Thx @rzo1!

mawiesne:
Once merged, could this be 🍒 picked to 9.1.x maintenance branch for a potential 9.1.4 release? @rzo1

rmannibucau:
Shouldnt it use injector if cdi is disabled and nothing else (owbinjector)?
Can mean scanning is wrongly configured for tests no?

rzo1:
> Shouldnt it use injector if cdi is disabled and nothing else (owbinjector)? Can mean scanning is wrongly configured for tests no?

We don't have a lot of tests with `@Application`and docs are sparse in that area, so might not get it correctly.

Adding the actual test class to `@Classes` of the `@Application` will inject the `@Resource` into the test otherwise it seems, that` @Resource` is not scanned anywhere else (and the fallbacks only inject `@Inject` fields).

Would that be the actual way to solve it with application composer? (If so, we just need to update the docs).

rmannibucau:
Test classes are injected and resources already handled.
Limit for single instance for suites is that classes must be scanned for injections to be prepared as any bean so test classes must be scanned.
Can be adding it in classes or marking test classes to be a cdi bean (`@Dependent`)

rzo1:
Thx @rmannibucau for clarification - I will update the test, verify it works and than remove my changes and update the docs, so it is clear how to use from user perspective.

mawiesne:
I can confirm that adding the test class `@Classes` helps, for example:

```
@Module
@Classes(cdi = true, value = {MyTest.class, App.class})
  public EjbJar modules() {
    return new EjbJar();
  }
```

Thx @rmannibucau for clarification. Thx @rzo1 for updating the docs - so community can better understand what is required in such a scenario.

rzo1:
Thx for checking it, @mawiesne and @rmannibucau for pointing out the limitations of the single instance in suites.

PR #1230, opened 2024-07-02
TOMEE-4357 - Add Jandex Index cache - 9.x

fataOT:
Add a Jandex index cache to TomEEMicroProfileListener

This will fix a issue with a longer startup time since TomEE 9

The processApplication method is called for every created application and will then create an index for all jars.
But the index is based on the AppInfo of the application. The AppInfo can have multiple applications.
So currently the same index is created for every single application which is unnecessary

This change will boost our start up time from 2min to ~45s

### DIFF
--- a/tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
+++ b/tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
@@ -19,6 +19,7 @@ package org.apache.tomee.microprofile;
 import io.smallrye.opentracing.SmallRyeTracingDynamicFeature;
 import jakarta.servlet.ServletContext;
 import jakarta.servlet.ServletRegistration;
+import org.apache.openejb.assembler.classic.AppInfo;
 import org.apache.openejb.assembler.classic.WebAppInfo;
 import org.apache.openejb.config.NewLoaderLogic;
 import org.apache.openejb.config.event.EnhanceScannableUrlsEvent;
@@ -43,10 +44,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.security.CodeSource;
-import java.util.Collection;
-import java.util.Enumeration;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.stream.Collectors;
@@ -65,6 +63,8 @@ public class TomEEMicroProfileListener {
         "io.smallrye.opentracing.contrib.interceptor.OpenTracingInterceptor",
         "io.smallrye.faulttolerance.FaultToleranceExtension",
         };
+
+    private static final Map<AppInfo, Index> indexCache = new HashMap<>();
 
     @SuppressWarnings("Duplicates")
     public void enhanceScannableUrls(@Observes final EnhanceScannableUrlsEvent enhanceScannableUrlsEvent) {
@@ -122,7 +122,13 @@ public class TomEEMicroProfileListener {
         // index only once and pass it everywhere it's needed. Also in order to build the index, we need the entire
         // application so doing it here from the AppInfo is way simpler
         try {
-            final Index index = of(afterApplicationCreated.getEvent().getApp().libs);
+            final Index index;
+            if(indexCache.containsKey(afterApplicationCreated.getEvent().getApp())) {
+                index = indexCache.get(afterApplicationCreated.getEvent().getApp());
+            } else {
+                index = of(afterApplicationCreated.getEvent().getApp().libs);
+                indexCache.put(afterApplicationCreated.getEvent().getApp(), index);
+            }
             MicroProfileOpenApiRegistration.registerOpenApiServlet(context, index);
 
         } catch (final IOException e) {
--- a/tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
+++ b/tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
@@ -21,6 +21,7 @@ import jakarta.servlet.ServletContext;
 import jakarta.servlet.ServletRegistration;
 import org.apache.openejb.assembler.classic.AppInfo;
 import org.apache.openejb.assembler.classic.WebAppInfo;
+import org.apache.openejb.assembler.classic.event.AssemblerBeforeApplicationDestroyed;
 import org.apache.openejb.config.NewLoaderLogic;
 import org.apache.openejb.config.event.EnhanceScannableUrlsEvent;
 import org.apache.openejb.loader.Files;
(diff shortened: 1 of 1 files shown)


tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
```diff
@@ -43,10 +44,7 @@
 import java.io.InputStream;
 import java.net.URL;
 import java.security.CodeSource;
-import java.util.Collection;
```
rmannibucau, line 46: IDE autoformat?

tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
```diff
@@ -122,7 +122,13 @@ public void processApplication(@Observes final BeforeEvent<AfterApplicationCreat
         // index only once and pass it everywhere it's needed. Also in order to build the index, we need the entire
         // application so doing it here from the AppInfo is way simpler
         try {
-            final Index index = of(afterApplicationCreated.getEvent().getApp().libs);
+            final Index index;
+            if(indexCache.containsKey(afterApplicationCreated.getEvent().getApp())) {
```
rmannibucau, line 126: you can do a get and test null which is always better than a double hash (contains + get)

tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
```diff
@@ -122,7 +122,13 @@ public void processApplication(@Observes final BeforeEvent<AfterApplicationCreat
...
+                index = indexCache.get(afterApplicationCreated.getEvent().getApp());
+            } else {
+                index = of(afterApplicationCreated.getEvent().getApp().libs);
+                indexCache.put(afterApplicationCreated.getEvent().getApp(), index);
```
rmannibucau, line 130: you likely want to listen for application undeployment and drop it from the cache cause like that it leaks in the app server

tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
```diff
@@ -66,6 +64,8 @@ public class TomEEMicroProfileListener {
         "io.smallrye.faulttolerance.FaultToleranceExtension",
         };
 
+    private static final Map<AppInfo, Index> indexCache = new HashMap<>();
```
rmannibucau, line 67: guess it can not be static since it belongs to the listener instance

rmannibucau (changes requested):

tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
fataOT: yes, fixed

tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
fataOT: I changed it to computeIfAbsent

tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
fataOT: removed static, it's indeed not needed

tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
fataOT: Ok, I need some help here, there is no method for this?

fataOT: I could use the AssemblerBeforeApplicationDestroyed event for this but actually the cache is only needed for creation. 
A better approach could be to remove the index from the cache, if the corresponding AppInfo is fully deployed.

Is there a event for this?

rmannibucau: I'd observe this destroyed event as a guarantee (error case?) but `AssemblerAfterApplicationCreated` for success case should be fine if it is not needed after assembler need

fataOT: AssemblerAfterApplicationCreated seems not to be the event I'm looking for, it is called before AfterApplicationCreated.

rmannibucau: depends the mode you are in (for webapps it is likely different than for ear).
one trivial mode can be to just add a cdi observer and do it after validation event (+ error case fallback as mentionned earlier)? can be simpler to just follow the lifecycle and use appfinder to lookup the appinfo

side note: i didnt check how relevant it is there (sorry if not but I tend to use another MP flavor) but we tend to attach runtime data (jandex there) to AppContext/webcontext objects this way no need to care of the destruction so it can be saner to add a flag in appinfo to flag it as requiring jandex and compute jandex lazily (in cdi builder?) from info tree data.
It would be stored in AppContext#data (or WebContext#data if relevant).
This way the lifecycle is easier to manage and it is more aligned on the internals but it is a bit more complex to wire.
Since CdiBuilder has both it can be a good place to create jandex if flagged as such from the observer you patched and to clean it up after it was built (so cdi is started, no more need of jandex I guess).

fataOT: I followed another approach. The AppInfo knows all webApps that need to be deployed.
I will remove the cache entry, if the last webApp for the given AppInfo has been deployed.

As suggested, I also listen for the AssemblerBeforeApplicationDestroyed event and will remove the corresponding cache entry, if it still exists

tomee/tomee-microprofile/mp-common/src/main/java/org/apache/tomee/microprofile/TomEEMicroProfileListener.java
```diff
@@ -66,6 +71,8 @@ public class TomEEMicroProfileListener {
         "io.smallrye.faulttolerance.FaultToleranceExtension",
         };
 
+    private final Map<AppInfo, Index> indexCache = new HashMap<>();
```
rmannibucau, line 74: should still be concurrent ;)

rest looks ok

fataOT: done 👍

rzo1 (approved):

PR #1278, opened 2024-07-11
[TOMEE-4355] Try to workaround Mojarra+OWB integration

jungm:
Mojarra 4.0.1+ doesn't run on OWB out of the box anymore, there is an ongoing CDI spec discussion happening in https://github.com/eclipse-ee4j/mojarra/issues/5457 that should resolve this again in the future

However, **right now** we have the following options:
- Downgrade to 4.0.0 and wait for an OWB/Mojarra release, which can take some time
- Write a workaround so that Mojarra 4.0.1+ can run again in TomEE

I've tried to workaround this by disabling Mojarra's CDI extension and enabling a subclass of it that lives in tomee-mojarra. This extension uses OWB internals to change the returnType of the beans Mojarra registers so that OWB proxies the correct types again. This fixes [TOMEE-4355](https://issues.apache.org/jira/browse/TOMEE-4355) without forcing us to stay on Mojarra 4.0.0


tomee/tomee-mojarra/src/main/resources/META-INF/openwebbeans/openwebbeans.properties
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+configuration.ordinal=75
+
+# We need to workaround an integration issue between Mojarra and OWB right now
+# For more details see https://github.com/apache/openwebbeans/pull/89
+# and https://github.com/eclipse-ee4j/mojarra/issues/5457
+org.apache.webbeans.ignoredExtensions=com.sun.faces.cdi.CdiExtension
```
rmannibucau, line 22: I'd rather not use it since the values are not merged so it means users can't rely on that anymore

tomee/tomee-mojarra/src/main/resources/META-INF/services/jakarta.enterprise.inject.spi.Extension
```diff
@@ -0,0 +1 @@
+org.apache.tomee.mojarra.owb.OwbCompatibleCdiExtension
```
rmannibucau, line 1: using the programmatic approach you can drop this file

tomee/tomee-mojarra/src/main/java/org/apache/tomee/mojarra/owb/MojarraAfterBeanDiscoveryDecorator.java
```diff
@@ -0,0 +1,76 @@
...
+import jakarta.enterprise.inject.spi.configurator.ObserverMethodConfigurator;
+import org.apache.webbeans.config.WebBeansContext;
+
+public class MojarraAfterBeanDiscoveryDecorator implements AfterBeanDiscovery {
+    private final AfterBeanDiscovery delegate;
+    private final WebBeansContext webBeansContext;
+
+    public MojarraAfterBeanDiscoveryDecorator(final AfterBeanDiscovery delegate) {
+        this.delegate = delegate;
+        this.webBeansContext = WebBeansContext.currentInstance();
```
rmannibucau, line 34: guess it is saner to pass the context programmatically so surely using a programmatic approach

it would also enable to replace the extension properly without having to potentially override user conf or this conf to be lost if user overrides it

I'd encourage you to maybe try something around https://github.com/apache/tomee/blob/e953d0665d207b730e3f21cc032f1cf75123b9c9/container/openejb-core/src/main/java/org/apache/openejb/cdi/OptimizedLoaderService.java#L86, in particular the `isFiltered` method which can replace the openwebbeans.properties in a better way IMHO.
Modifying a little bit the logic you can also switch the extension:


```
        while (it.hasNext()) {
            if (it.hasNext()) {
                if (isFiltered(extensionCopy, it.next())) {
                    it.remove();
                }
            }
        }
```

->

```
        List<Extension> toAdd = null;
        while (it.hasNext()) {
            var next = it.next();
            var replacement = replace(next);
            if (replacement != null && !isFiltered(extensionCopy, replacement)) {
                (toAdd == null ? toAdd = new ArrayList<>()  : toAdd).add(replacement);
            }
            if (isFiltered(extensionCopy, next)) {
                it.remove();
            }
        }
        list.addAdd(toAdd);

....

    public Extension replace(final Extension current) { // + try/catch(e){/*ignore*/}
        return Objects.equals("....CdiExtension", current.getClass().getName()) ? current.getClass().getClassLoader().loadClass("....replacment").asSubClass(Extension.class).getConstructor().newInstance();
    }
```

(in the spirit)

rmannibucau (commented):
proposed another way not taking user configuration toggles

tomee/tomee-mojarra/src/main/java/org/apache/tomee/mojarra/owb/MojarraAfterBeanDiscoveryDecorator.java
jungm: Rewrote this part and also how the extension is replaced/registered, actually wasn't aware that OWB does not merge ignoredExtensions between different configs

I access the current WebBeansContext using reflection from the InjectableBeanManager now but that at least seems a bit unintuitive. Maybe there's an easier way I just don't know of?

rmannibucau: think i would go to add to the loader service a `setWebBeansContext(WBC)` - every time it is instantiated it has the WBC a few line after so calling the setter should be easy.
this way in replace you can just new MyExtension(wbc) or alike.

jungm: done

tomee/tomee-mojarra/src/main/java/org/apache/tomee/mojarra/TomEEMojarraService.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.openejb.loader.SystemInstance;
+import org.apache.openejb.observer.Observes;
+import org.apache.openejb.spi.Service;
+import org.apache.tomee.mojarra.owb.OwbCompatibleCdiExtension;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+public class TomEEMojarraService implements Service {
```
rmannibucau, line 31: I'm not strictly opposed to that but if you set a threadlocal you should reset it

my 2 cts: drop the service, drop the threadlocal and just hardcode the extension replacement in the loader, this is a global replacement not per instance

jungm: done

rmannibucau (approved):
LGTM

PR #1328, opened 2024-07-27
TOMEE-4375 - Fix NPEs in SavedRequest

mpuening:


mpuening:
Stack trace fetching "Content-Type" header:

 java.lang.NullPointerException: Cannot invoke "java.util.List.get(int)" because the return value of "java.util.Map.get(Object)" is null
	org.apache.tomee.security.http.SavedRequest$1.getHeader(SavedRequest.java:110)
	jakarta.servlet.http.HttpServletRequestWrapper.getHeader(HttpServletRequestWrapper.java:83)
	com.sun.faces.context.RequestHeaderMap.get(RequestHeaderMap.java:48)
	com.sun.faces.context.RequestHeaderMap.get(RequestHeaderMap.java:32)
	java.base/java.util.Collections$UnmodifiableMap.get(Collections.java:1656)
	jakarta.faces.application.ViewHandler.calculateCharacterEncoding(ViewHandler.java:363)
	jakarta.faces.application.ViewHandler.initView(ViewHandler.java:236)
	com.sun.faces.application.view.MultiViewHandler.initView(MultiViewHandler.java:110)
	jakarta.faces.application.ViewHandlerWrapper.initView(ViewHandlerWrapper.java:81)
	com.sun.faces.lifecycle.RestoreViewPhase.doPhase(RestoreViewPhase.java:95)
	com.sun.faces.lifecycle.LifecycleImpl.execute(LifecycleImpl.java:131)
	jakarta.faces.webapp.FacesServlet.executeLifecyle(FacesServlet.java:691)
	jakarta.faces.webapp.FacesServlet.service(FacesServlet.java:449)
	org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:51)
	org.apache.openejb.server.httpd.EEFilter.doFilter(EEFilter.java:67)

tomee/tomee-security/src/main/java/org/apache/tomee/security/http/SavedRequest.java
```diff
@@ -102,12 +102,20 @@ public Enumeration<String> getHeaderNames() {
...
+                }
+                return Collections.enumeration(header);
             }
 
             @Override
             public String getHeader(String name) {
-                return headers.get(name).get(0);
+                List<String> header = headers.get(name);
+                if (header == null || header.isEmpty()) {
+                    return null;
```
jungm, line 116: Not sure if we should turn an empty header value into null implicitly, I'll check later what other implementations of HttpServletRequest do in that case

jungm: Servlet 6.0 spec says:
> Returns the value of the specified request header as a String. If the request did not include a header of the specified name, this method returns null. If there are multiple headers with the same name, this method returns the first head in the request. The header name is case insensitive. You can use this method with any request header.

IMO this means we must not alter the header value if it is present. From a quick glance this is also how tomcat/jetty work

tomee/tomee-security/src/main/java/org/apache/tomee/security/http/SavedRequest.java
```diff
@@ -102,12 +102,20 @@ public Enumeration<String> getHeaderNames() {
...
+                if (header == null) {
+                    header = Collections.emptyList();
+                }
+                return Collections.enumeration(header);
             }
 
             @Override
             public String getHeader(String name) {
-                return headers.get(name).get(0);
+                List<String> header = headers.get(name);
```
jungm, line 114: in `getHeaders(String)` you assume `headers`may be null, this check is missing here

jungm (changes requested):
Nice PR!
I've commented a few minor details I would still like to see changed. Also `getHeaderNames()` is not included in your changes

jungm:
@mpuening can you fix `getHeaderNames()` as well? Then it should look good I think

jungm (approved):

rzo1 (approved):

PR #1485, opened 2024-09-10
TOMEE-4404 - fix broken tomee-webapp

lherschi:
TomcatClassPath.java: fix call of a wrong method and thus the resulting NPE
TomcatWebAppBuilder.java: fix NPE
war.xml: fix non-functioning display of the tomee installer webapp

### DIFF
--- a/container/openejb-loader/src/main/java/org/apache/openejb/loader/TomcatClassPath.java
+++ b/container/openejb-loader/src/main/java/org/apache/openejb/loader/TomcatClassPath.java
@@ -153,7 +153,7 @@ public class TomcatClassPath extends BasicURLClassPath {
                 try {
                     final Object cp = getURLClassPath((URLClassLoader) getClassLoader());
                     final Class<?> clazz = cp.getClass();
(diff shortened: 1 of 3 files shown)


lherschi:
The Tomee web app will also be available for download for the last time with version 8.0.6. This was followed by a change in the naming of the git tags from tomee-8.0.6 to tomee-project-x.x.x. This may simply mean that it is no longer built automatically.

jungm:
At a first glance your changes look good to me, kudos! The reason for the webapp not being published anymore after 8.0.6 is that the TomEE project actually decided to retire these some time ago for various reasons, see this discussion: https://lists.apache.org/thread/31w7336d5ycqhmoy4pngbsjg3odm0yqx

Your PR makes me think you're upgrading from a TomEE version from before 8.0.8 to something more recent while using Tomcat with the TomEE webapp - I'd highly suggest you to switch to the full TomEE server bundles that are available to download on https://tomee.apache.org/download-archive.html

jungm:
Please note the TomEE Webapp is retired and can be fully removed from source code at any time

PR #1509, opened 2024-09-23
TOMEE-2324 - Remove Corba ORB APIs from Codebase

rzo1:
The CORBA part is optional in EE10. We discussed this issue on the list a while back (TomEE 8.0.5 / 8.0.6) and there was consensus to remove CORBA in a new major version of TomEE.

We missed that for 9.1.x but since 10.x is not GA yet, it might be a good moment to remove it. This will remove the need to shade servicemix-bcel and ship yoko in TomEE, which is also quite outdated and unmaintained.


jungm (approved):
Not familiar with ORB/CORBA at all to be fair, but the diff does look good to me

mawiesne (approved):

PR #1996, opened 2025-06-30
Threadsafe RequestScopedThreadContextListener

ammannmi:
Storing the DestroyContext in the ThreadContext itself is not threadsafe if the context is propagated to another thread (this is done in the ApplicationThreadContextProvider). This can occure in ManagedExecutors. Dependant on the timing the main thread could remove the RequestScope unfortunally.

### DIFF
--- a/container/openejb-core/src/main/java/org/apache/openejb/cdi/RequestScopedThreadContextListener.java
+++ b/container/openejb-core/src/main/java/org/apache/openejb/cdi/RequestScopedThreadContextListener.java
@@ -30,6 +30,8 @@ import jakarta.enterprise.context.spi.Context;
  * @version $Rev$ $Date$
  */
 public class RequestScopedThreadContextListener implements ThreadContextListener {
+    private final static ThreadLocal<DestroyContext> destroyable = new ThreadLocal<>();
+    
     @Override
     public void contextEntered(final ThreadContext oldContext, final ThreadContext newContext, boolean propagateTx) {
 
@@ -46,7 +48,7 @@ public class RequestScopedThreadContextListener implements ThreadContextListener
 
         if (requestContext == null) {
             contextsService.startContext(RequestScoped.class, CdiAppContextsService.EJB_REQUEST_EVENT);
-            newContext.set(DestroyContext.class, new DestroyContext(contextsService, newContext));
+            destroyable.set(new DestroyContext(contextsService, newContext));
         }
     }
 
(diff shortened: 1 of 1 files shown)


otbutz:
Possibly related: https://issues.apache.org/jira/browse/TOMEE-4485

rzo1 (approved):

rzo1:
Can we try to come up with a test case for this? Might be difficult to hit because of timing, but maybe worth a try?

ammannmi:
The problem mentioned in https://issues.apache.org/jira/browse/TOMEE-4485 with the missing ClassLoader was the primary reason for the analysis. When ManagedExecutors where used for some asynchronous tasks, the error occured in the main (request) thread, dependant on the duration of the asynchronous tasks and the remaining time in the main thread.

dblevins:
Hey All,

First some insights on the design of ThreadContext and TheadContextListener.  It was essentially created to be the "one thread local to rule them all" and eliminate the need for multiple thread locals.  A ThreadContext is supposed to be thread-safe and not shared by multiple threads.  On the surface with that design in mind, there should be no reason to have a ThreadLocal in a ThreadContextListener as is done on this PR.  That's essentially putting a ThreadLocal on a ThreadLocal.  If handling of the first thread local is broken and the ThreadContext is getting leaked, we'd want to fix that issue as it would be very serious.  Said another way, we'd want to fix the first TheadLocal handling vs leaving it broken and adding more TheadLocals.

@ammannmi @otbutz could you both subscribe to the dev list so we can talk this through there? [Subscribe](mailto:dev-subscribe@tomee.apache.org).  I'll wait till tomorrow before starting a thread there so you have time to subscribe.

otbutz:
> @ammannmi @otbutz could you both subscribe to the dev list so we can talk this through there? [Subscribe](mailto:dev-subscribe@tomee.apache.org). I'll wait till tomorrow before starting a thread there so you have time to subscribe.

done

ammannmi:
I aggree with @rzo1, the ThreadContext should be bound to one thread and not shared between threads because this would require, that the whole context is threadsafe. So the main reason for the problems is sharing the ThreadContext.

ammannmi:
I subscribed to the list, I hope I did it correctly

rzo1:
The thread is here https://lists.apache.org/thread/9hhxsrclzhc49ln74xc029j836d6g7zq

PR #2033, opened 2025-08-02
TOMEE-4518 - remove transaction propagation

jungm:
See [TOMEE-4518](https://issues.apache.org/jira/browse/TOMEE-4518) and [this thread](https://lists.apache.org/list.html?dev@tomee.apache.org) on the dev list

Maybe we should think out loud about failing the deployment of a `ContextService` if it defines `Propagated*=Transaction`. This however brings up questions again like what to do in this case:
```
<Resource id="Foo" type="ContextService">
  Propagated = Remaining
  Cleared =
  Unchanged =
</Resource>
```
And if we should completely abort the TomEE startup if an illegal ContextService is detected in `tomee.xml`. What if it has `Lazy=true`? So I chose to not enforce this via code yet.


jgallimore:
This looks good to me. I've rebased it. +1 to merge.

PR #2577, opened 2026-04-03
Concurrency 3.1

rzo1:
First version - think we can improve or re-work if needed.


container/openejb-core/src/main/java/org/apache/openejb/cdi/concurrency/AsynchronousInterceptor.java
```diff
@@ -88,26 +117,225 @@ public Object aroundInvoke(final InvocationContext ctx) throws Exception {
         return ctx.getMethod().getReturnType() == Void.TYPE ? null : future;
     }
 
+    private Object aroundInvokeScheduled(final InvocationContext ctx, final Asynchronous asynchronous,
+                                          final Schedule[] schedules) throws Exception {
+        // Per spec, the executor attribute may reference either a ManagedScheduledExecutorService
+        // or a plain ManagedExecutorService. When a plain MES is referenced, fall back to the
+        // default MSES for scheduling capability but preserve the MES's context service.
+        final ManagedScheduledExecutorServiceImpl mses = resolveMses(asynchronous.executor());
```
jungm, line 125: interestingly enough concurro completely disregargs the configured executor, claiming it is not to be used in the scheduled case according to the spec (though i cannot find where this behavior is described)
https://github.com/eclipse-ee4j/glassfish-concurro/blob/3.1.0/src/main/java/org/glassfish/concurro/cdi/asynchronous/AsynchronousInterceptor.java#L127-L132

openliberty seems to at least honor the contextservice that is attached to the used executor

container/openejb-core/src/main/java/org/apache/openejb/cdi/concurrency/AsynchronousInterceptor.java
```diff
@@ -88,26 +117,225 @@ public Object aroundInvoke(final InvocationContext ctx) throws Exception {
...
+        //   - the method raises an exception
+        //   - the future is completed (via Asynchronous.Result.complete()) or cancelled
+        final CompletableFuture<Object> outerFuture = mses.newIncompleteFuture();
+        final AtomicReference<ScheduledFuture<?>> scheduledRef = new AtomicReference<>();
+        final AtomicReference<LastExecution> lastExecutionRef = new AtomicReference<>();
+
+        // Extract method and target from InvocationContext for direct invocation.
+        // We must NOT reuse ctx.proceed() — InvocationContext's interceptor iterator
+        // is single-use, so subsequent proceed() calls would bypass TX/security interceptors.
+        final Method beanMethod = ctx.getMethod();
```
jungm, line 153: this is likely not fixing what it is trying to fix as this returns the raw method and not the method on the CDI proxy that would run interceptors, pushed a test to prove this regression and working on a fix

jungm (approved):

PR #2579, opened 2026-04-04
Security 4.0 - TCK Setup

rzo1:
First step: TCK Setup is working (does not run in CI).


jungm:
note: 3.0 TCK had some weird quirks in the past, see https://github.com/apache/tomee-tck/commit/a4566ffe0b215554e69e096bc1081acdb6650a63

rzo1:
> note: 3.0 TCK had some weird quirks in the past, see [apache/tomee-tck@a4566ff](https://github.com/apache/tomee-tck/commit/a4566ffe0b215554e69e096bc1081acdb6650a63)

Yes. This might be needed to fix some of the current TCK failures. Currently, the setup is capable of running the TCK and we have some green stuff (as expected). Guess we can move forward with it, so we can start with the security impl :)

PR #2619, opened 2026-04-16
TOMEE-4600 - remove openejb.cache.cdi-type-resolution-failure

jungm:
see https://issues.apache.org/jira/browse/TOMEE-4600


rzo1 (approved):
I think @tandraschko has handled that in the current OWB snapshot, so ok for me.

PR #2658, opened 2026-05-05
TOMEE-4604 strip ssl keystore configuration params from outgoing URL.

ksmigrod:
Sorry for the refactor of HttpConnectionTest, but it was hard to understand why it fails without separators between parts of response.

### DIFF
--- a/server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
+++ b/server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
@@ -151,6 +151,7 @@ public class HttpConnectionFactory implements ConnectionFactory {
             }
 
             final StringBuilder sb = new StringBuilder();
+            // split on [&?] to account for failover URLs with multiple http URLs embedded
             for (String param : url.substring(queryStartIndex+1).split("[&?]")) {
                 final int p = param.indexOf('=');
                 if ((p < 0 && Arrays.binarySearch(params, param) < 0)
--- a/server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
+++ b/server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
@@ -155,7 +155,7 @@ public class HttpConnectionFactory implements ConnectionFactory {
             for (String param : url.substring(queryStartIndex+1).split("[&?]")) {
                 final int p = param.indexOf('=');
                 if ((p < 0 && Arrays.binarySearch(params, param) < 0)
-                        || (p > 0 && Arrays.binarySearch(params, param.substring(0, p)) < 0)) {
+                        || (p >= 0 && Arrays.binarySearch(params, param.substring(0, p)) < 0)) {
                     if (!sb.isEmpty()) { sb.append('&'); }
                     sb.append(param);
                 }
--- a/server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
+++ b/server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
@@ -60,7 +60,7 @@ public class HttpConnectionFactory implements ConnectionFactory {
     }
 
     public static class HttpConnection implements Connection {
-        private final static String []PARAMS_TO_STRIP = new String[]{
+        private static final String[] PARAMS_TO_STRIP = new String[]{
                 // must be sorted alphabetically
                 "authorization", "authorizationHeader", "basic.password", "basic.username",
                 "connectTimeout", "readTimeout",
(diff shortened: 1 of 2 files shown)


server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
```diff
@@ -59,6 +60,13 @@ public Connection getConnection(final URI uri) throws IOException {
     }
 
     public static class HttpConnection implements Connection {
+        private final static String []PARAMS_TO_STRIP = new String[]{
```
rzo1: Style nit (carried over from #2657): Apache/openejb-client convention is `static final` (not `final static`) and `String[] NAME` (not `String []NAME`).

```suggestion
        private static final String[] PARAMS_TO_STRIP = new String[]{
```

server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
```diff
@@ -140,20 +141,21 @@ public HttpConnection(final URI uri, final ConcurrentMap<URI, SSLSocketFactory>
...
         }
 
-        private String stripQuery(final String url, final String param) {
-            String result = url;
-            do {
-                final int h = result.indexOf(param + '=');
-                int end = result.indexOf('&', h);
-                if (end < 0) {
-                    end = result.length();
+        private String stripQuery(final String url, final String []params) {
```
rzo1: Same `String []params` -> `String[] params`. Also, the brace-on-same-line one-liners on the next two lines are out of style with the rest of this file (every other early-return uses a multi-line block).

server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
```diff
@@ -140,20 +141,21 @@ public HttpConnection(final URI uri, final ConcurrentMap<URI, SSLSocketFactory>
...
+            final int queryStartIndex = url.indexOf('?');
+            if (queryStartIndex < 0) { return url; }
+            if (queryStartIndex + 1 == url.length()) { return url.substring(0, queryStartIndex); }
+
+            final StringBuilder sb = new StringBuilder();
+            for (String param : url.substring(queryStartIndex+1).split("[&?]")) {
```
rzo1, line 155: Worth a one-line comment here explaining why the regex is `[&?]` and not just `&`: it's load-bearing for the failover URI case (`complexURIAuthorization`) where the chained URI ends up with an embedded `?...` after the first param is stripped. Otherwise a future reader will simplify it back to `"&"` and break that test.

server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
```diff
@@ -140,20 +141,21 @@ public HttpConnection(final URI uri, final ConcurrentMap<URI, SSLSocketFactory>
...
+                final int p = param.indexOf('=');
+                if ((p < 0 && Arrays.binarySearch(params, param) < 0)
```
rzo1, line 157: Edge case still open from the previous review: a token shaped `=value` (empty name, `p == 0`) matches neither branch and is silently dropped. `p >= 0` would be more defensible than `p > 0`. Won't happen with well-formed URIs, just brittleness.

rzo1 (commented):
Both blocking issues from #2657 are resolved (the `Arrays` import and the unconditional use of `newUrl`), and the new `httpStripsSSLQueryParams` test directly proves the SSL-only leak is closed. Just a few small follow-ups inline.

server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
ksmigrod: [40fed5a](https://github.com/apache/tomee/pull/2658/commits/40fed5a60931494b25e7ffb749184a87797594f9)

server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
ksmigrod: [c102a25](https://github.com/apache/tomee/pull/2658/commits/c102a25f2e240fc233e75f27e01dc0254d554e76)

server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
ksmigrod: [658772c](https://github.com/apache/tomee/pull/2658/commits/658772ced150dc62f3915f9227cffd1b700bfbde)

server/openejb-client/src/main/java/org/apache/openejb/client/HttpConnectionFactory.java
ksmigrod: [40fed5a](https://github.com/apache/tomee/pull/2658/commits/40fed5a60931494b25e7ffb749184a87797594f9)

rzo1:
@ksmigrod Thanks. If you need it also for 10.x - can you open a PR for 10.x too with the cherry pick?

PR #2856, opened 2026-07-28
TOMEE-4662 - Speed up annotation processing during WAR deploy

tandraschko:
## Speed up annotation processing during WAR deploy
`OpenEJBContextConfig.isIncluded()` resolves the canonical path of the module
roots and of the web resource on every call, and Tomcat calls
`processAnnotationsWebResource()` once per web resource. On a WAR with a few
hundred classes under `WEB-INF/classes` that is thousands of filesystem
syscalls, most of them resolving the same handful of paths again and again.
Three changes, no behaviour change:
- Memoise `File.getCanonicalFile()` and the URL to `File` conversion. Both
  mappings are stable for the lifetime of a deploy and the config instance is
  per context.
- Return early from `processAnnotationsWebResource()` once every module has
  been processed. There is nothing left to match the resource against, and that
  branch never delegates to `super`, so resolving the resource is wasted work.
- Reuse the path from `FileResource.getCanonicalPath()` instead of letting
  `isIncluded()` resolve the same, already canonical, path a second time.
Measured on a synthetic WAR (40 jars, 513 classes in `WEB-INF/classes`) with
`tomee-embedded`: `isIncluded()` accounted for 165 of 468 deploy-phase profiler
samples before the change and 1 after, and deploy wall time dropped ~25%.
Note this was measured on Windows, where `getCanonicalPath()` is considerably
more expensive than on Linux, so expect a smaller gain there.

Also see https://github.com/apache/tomcat/pull/1030

### DIFF
--- a/tomee/tomee-catalina/src/main/java/org/apache/catalina/startup/OpenEJBContextConfig.java
+++ b/tomee/tomee-catalina/src/main/java/org/apache/catalina/startup/OpenEJBContextConfig.java
@@ -89,6 +89,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class OpenEJBContextConfig extends ContextConfig {
     private static Logger logger = Logger.getInstance(LogCategory.OPENEJB, OpenEJBContextConfig.class);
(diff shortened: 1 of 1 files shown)


jungm:
https://ci-builds.apache.org/job/Tomee/view/tomee-11.x/job/pull-request-manual-with-tck/52/

jungm:
Nice speedup, logic checks out. Two changes I'd make before merge:

1. **Use `ConcurrentHashMap` for both caches.** With `parallelAnnotationScanning="true"`, Tomcat runs `processAnnotationsUrl` on multiple utility-executor threads, which reaches the maps via `internalProcessAnnotations` → `isIncludedIn`. Concurrent puts into a plain `HashMap` can lose entries or corrupt the table.

2. **Clear both maps in `configureStop()`**, like `webInfClassesAnnotationsProcessed`. The listener survives context stop/start, so after a reload `canonicalFiles` can hold stale resolutions (symlinks/paths may have changed), and both maps otherwise retain ~one entry per class file for the life of the context.

Also: please file a TOMEE JIRA and reference it in the title/commit, per project convention.

tandraschko:
both done: https://issues.apache.org/jira/browse/TOMEE-4662

jungm (approved):

rzo1 (approved):

PR #2895, opened 2026-08-02
TOMEE-4673 - Remove "About the 'tomee' webapp" since Tomee drop-in webapp was removed in TOMEE-3724

Sxtormulo:
Remove the documentation abouth the drop-in Tomee webapp as this was removed in Jira [TOMEE-3724](https://issues.apache.org/jira/browse/TOMEE-3724). And modify the documentation page to remove the reference for the [tomee-webpage](https://issues.apache.org/jira/browse/TOMEE-webpage).adoc.


jungm:
thx for your contribution! two more places I found where the webapp is still mentioned in docs would be:
- installation-drop-in-war.adoc -> would remove that and the link to it entirely
- index.adoc in docs/developer/testing/arquillian still mentions webapp

would you also update that? then I'm happy to merge :slightly_smiling_face:

Sxtormulo:
Understood. Thank you very much; I hadn’t realised it was mentioned there as well.
I used grep to look for other mentions of ‘tomee drop-in webapp’ and found a few more.
I also found that there are sections where the `tomee.remote.support` property is mentioned, which refers to the `/tomee webapp`, such as at https://tomee.apache.org/tomee-11.0/docs/admin/configuration/server. Should I remove this mention as well?

jungm:
from what I can tell right now this is something entirely different, the tomee-webapp we "dropped" (really just unsupported, it still exists in our code) was modifying tomcat and injecting a bunch of Jakarta EE implementations to turn a stock Tomcat into TomEE

That property you found (tomee.remote.support) seems to register a servlet which exposes ejbs, so really something completely different that I wouldn't touch in this PR

Sxtormulo:
I understand; I’ll focus on the ones relating to the drop-in. I’ll get on with it after work.

Sxtormulo:
Hello, good night.
I make the changes removing the additional references to tomee drop-in webapp.
:)

PR #2898, opened 2026-08-06
TOMEE-4674 include opentelemetry-sdk-extension-declarative-config and…

jgallimore:
… add itest.

### DIFF
--- a/itests/tomee-microprofile-itests/pom.xml
+++ b/itests/tomee-microprofile-itests/pom.xml
@@ -30,6 +30,11 @@
   <packaging>jar</packaging>
   <name>TomEE :: iTests :: TomEE MicroProfile</name>
 
+  <properties>
+    <opentelemetry.version>1.62.0</opentelemetry.version>
+    <opentelemetry.instrumentation.version>2.27.0</opentelemetry.instrumentation.version>
+  </properties>
+
   <build>
     <plugins>
       <plugin>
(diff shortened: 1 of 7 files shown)


itests/tomee-microprofile-itests/pom.xml
```diff
@@ -30,6 +30,11 @@
   <packaging>jar</packaging>
   <name>TomEE :: iTests :: TomEE MicroProfile</name>
 
+  <properties>
+    <opentelemetry.version>1.62.0</opentelemetry.version>
```
rzo1: Can we put (and manage) them in the root pom of TomEE?

tomee/tomee-microprofile/mp-common/pom.xml
```diff
@@ -292,6 +292,12 @@
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-extension-declarative-config</artifactId>
+      <version>1.62.0-alpha</version>
```
rzo1: Same as the comment above? Put it into root pom and dep manage it?

jgallimore:
@rzo1 Thanks for the review! I'll make those changes.

rzo1 (approved):

jungm (approved):

PR #2902, opened 2026-08-19
TOMEE-4682 TomEESecurityContext.authenticate() should delegate to Htt…

jgallimore:
…pServletRequest as opposed to invoking JASPIC directly. Includes 2 new examples demonstrating 2FA using passkeys.

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -206,6 +206,8 @@
     <module>security-tomcat-user-identitystore</module>
     <module>security-custom-identitystore</module>
     <module>security-openid</module>
+    <module>security-passkey-2fa</module>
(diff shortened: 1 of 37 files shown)


exabrial:
Hey Jonathan! Excellent patch! I did some local testing and found a couple of areas that are tangentially related:

* `@LoginToContinue` produces sessions 100% of the time, when it should really just produce one when redirecting to login. So if accessing a non-protected resource -> don't create a session. If accessing one -> create a session to save the request

* Speaking of the request; the saved request to be replayed isn't getting fully replayed.

I'll get these pushed to an additional PR in a few mins

exabrial:
https://github.com/jgallimore/tomee/pull/12/changes

PR #2939, opened 2026-09-08
TOMEE-4699 Fix ConcurrentModificationException in ThreadContext constructor

JorrenH:
See Jira ticket for full context.

In summary:
* HashMap#putAll on a synchronizedMap is *not* thread-safe. It will access `.entrySet().iterator()` internally, allowing it to cause a `ConcurrentModificationException`.
* This PR simply synchronizes on the data so the iterator call remains safe.

*Note*: ConcurrentHashMap does not work here because it does not allow `null` values.


otbutz:
> * HashMap#putAll on a synchronizedMap is _not_ thread-safe. It will access `.entrySet().iterator()` internally, allowing it to cause a `ConcurrentModificationException`.

To be precise, there are two mutexes involved: `putAll` only synchronises on `this.data`. However, accessing `that.data` via an iterator happens without synchronisation, which violates the contract of `Collections.synchronizedMap()`.

container/openejb-core/src/main/java/org/apache/openejb/core/ThreadContext.java
```diff
@@ -237,9 +239,15 @@ public String toString() {
     }
 
     private String dataToString(final Map<Class, Object> data) {
-        return data.entrySet().stream()
+        // copy data under monitor (synchronized map), format outside lock
+        return synchronizedCopy(data).entrySet().stream()
                 .map(entry -> entry.getKey() + "=" + (entry.getValue() == null ? "null" : entry.getValue().hashCode()))
                 .collect(Collectors.joining(", "));
+    }
```
otbutz, line 249:
```suggestion
        synchronized (data) {
                return data.entrySet().stream()
                        .map(entry -> entry.getKey() + "=" + (entry.getValue() == null ? "null" : entry.getValue().hashCode()))
                        .collect(Collectors.joining(", "));
        }
    }
```

otbutz (changes requested):
I see no reason to copy the whole map?

container/openejb-core/src/main/java/org/apache/openejb/core/ThreadContext.java
JorrenH: It was suggested on the mailing list, since arbitrary user code can run inside `.hashCode()`. Difficult for me to assess the trade-off on this.
> To keep the critical section short and avoid calling entry.getValue().hashCode() (arbitrary application code) while ho-lding a monitor that sits on the EJB invocation hot path, better to copy under the lock and format outside it.

rzo1: Keeping the copy is the safer option here (imho): that monitor is the same one `get`/`set`/`remove` take, and `InterceptorStack` takes it around every business method, so holding it across `entry.getValue().hashCode()` puts code we don't control on the EJB invocation path. No current value in that map does anything expensive in `hashCode()`, so this is about not depending on that, and formatting outside the monitor only costs one small `HashMap` copy on a debug-only path.

`toString()` reads `data.size()` outside the lock and then `dataToString` locks again, so the printed count and the printed entries come from two different reads. Folding the size into the same copy fixes that:

```java
private String dataToString(final Map<Class, Object> data) {
    final Map<Class, Object> copy = synchronizedCopy(data);
    return "(" + copy.size() + ")=" + copy.entrySet().stream()
            .map(entry -> entry.getKey() + "=" + (entry.getValue() == null ? "null" : entry.getValue().hashCode()))
            .collect(Collectors.joining(", "));
}
```

with `toString()` using `", data=" + dataToString(data)`.

rzo1 (approved):
LGTM, thanks for tracking this down.

I pushed a reproducer to the branch: `ThreadContextCopyRaceTest` copies a ThreadContext on one thread while another updates it. It fails within about half a second without the synchronized block and passes with it, so it should be a lot easier to work with than reproducing under load with a debugger attached. Feel free to change or drop it if you would rather keep the PR minimal.

I left one non-blocking note on the dataToString thread about `toString()` reading the size outside the lock.

JorrenH:
@rzo1 Good to have a quick reproducer, thanks! I updated toString() such that the size is consistent with the entries since it's a quick win. Did pull the whole line into `dataToString` to avoid a change in the format (`data=(2)=..` vs `data(2)=..`)

PR #2940, opened 2026-09-09
TOMEE-4703 / TOMEE-4704 capture thread context on the submitting thread

rzo1:
## TOMEE-4703 / TOMEE-4704 — capture thread context on the submitting thread

Follow-up to #2939, which fixes the CME itself and should still go in for 10.x. This targets `main`
and addresses why the race exists.

`ApplicationThreadContextProvider.currentContext()` runs on the submitting thread but stores a live
reference to its `ThreadContext`, and copies it later in `begin()`, on the thread running the task.

The SPI javadoc asks for the opposite: `ThreadContextProvider.currentContext` "Captures from the
current thread a snapshot of the provided thread context type", returning an "immutable snapshot
... captured from the current thread", and `ThreadContextSnapshot` "can be applied to any number of
threads, including concurrently". Spec section 4.1.1 puts the same duty on the provider: "Capture or
produce snapshots of the provided type of thread context when ThreadContextProvider methods are
invoked."

### Commits

1. `340376ed25` — `endContext()` restored the class loader before `ThreadContext.exit()`, which then
   overwrote it, leaving pool threads on the application class loader. Pre-existing, standalone,
   backport candidate for 10.3.0.
2. `c940d73042` — capture eagerly into an immutable `ThreadContext.Capture`; build a new
   `ThreadContext` from it per `begin()`. Excludes `InvocationContext` and `DestroyContext`, which are
   tied to the submitter's invocation.
3. `c6cb5080c9` — `currentContextExecutor()` captured in `execute()` instead of at creation, and one
   `CUTask.Context` per task broke concurrent use of a contextual proxy.
4. `9d79f7de06` — TOMEE-4704: `CUTask.invoke()` did its setup outside the `try`, so a failure there
   skipped the listener callbacks and the cleanup.

TCK run: https://ci-builds.apache.org/job/TomEe/job/pull-request-manual-with-tck/59/


rzo1:
@otbutz and @JorrenH : This is a 1st fix attempt - additional 👁️ 👁️  are always welcome.

container/openejb-core/src/main/java/org/apache/openejb/core/ThreadContext.java
```diff
@@ -151,13 +152,98 @@ public ThreadContext(final BeanContext beanContext, final Object primaryKey, fin
...
+            this.beanContext = that.beanContext;
+            this.primaryKey = that.primaryKey;
+
+            final Map<Class, Object> copy = new HashMap<>();
+            synchronized (that.data) {
+                for (final Map.Entry<Class, Object> entry : that.data.entrySet()) {
+                    if (!NON_PROPAGATED.contains(entry.getKey().getName())) {
+                        copy.put(entry.getKey(), entry.getValue());
+                    }
+                }
```
otbutz, line 223: Nitpick, but I would avoid the double inversion here.

```suggestion
                for (final Map.Entry<Class, Object> entry : that.data.entrySet()) {
                    if (NON_PROPAGATED.contains(entry.getKey().getName())) {
                        continue;
                    }
                    copy.put(entry.getKey(), entry.getValue());
                }
```

container/openejb-core/src/main/java/org/apache/openejb/core/ThreadContext.java
```diff
@@ -151,13 +152,98 @@ public ThreadContext(final BeanContext beanContext, final Object primaryKey, fin
...
+    /**
+     * Copy constructor. Must be called on the thread that owns <code>that</code>, since a
+     * ThreadContext is confined to its thread. Use {@link #capture()} to pass a context to
+     * another thread.
+     */
     public ThreadContext(final ThreadContext that) {
         this.beanContext = that.beanContext;
         this.primaryKey = that.primaryKey;
-        this.data.putAll(that.data);
+        synchronized (that.data) {
```
JorrenH: I would argue this rewrite removes the need for synchronization. If a single thread fully owns the ThreadContext, there cannot be any concurrent reads/writes.

container/openejb-core/src/main/java/org/apache/openejb/core/ThreadContext.java
```diff
@@ -151,13 +152,98 @@ public ThreadContext(final BeanContext beanContext, final Object primaryKey, fin
...
+        private final BeanContext beanContext;
+        private final Object primaryKey;
+        private final Map<Class, Object> data;
+
+        private Capture(final ThreadContext that) {
+            this.beanContext = that.beanContext;
+            this.primaryKey = that.primaryKey;
+
+            final Map<Class, Object> copy = new HashMap<>();
+            synchronized (that.data) {
```
JorrenH: Similarly the capture must be on the owning thread, so it should not need synchronization.

JorrenH (commented):
Fixes the issues and moves the implementation closer to spec which is good.

One question I had; is synchronization still needed when the capture is synchronous on the owning thread? I.e. is it still possible for other threads to directly access the ThreadContext with these changes?

rzo1:
@JorrenH Good question. For the paths this PR changes, no. `capture()` runs on the thread that owns the context, `begin()` builds a new `ThreadContext` per task on the thread running it, and `SecurityThreadContextProvider` only copies the context that was just entered on that same thread. So the concurrency utilities no longer touch another thread's `ThreadContext`.

I'd still keep the `synchronized` blocks for now:
- `data` is still a `Collections.synchronizedMap`, and its contract requires holding the map's monitor while iterating. Dropping the blocks but keeping the map would just be incorrect in a different way.
- `ThreadContext` is public, and references to it escape in other places. For example, `PoolEndpointHandler` keeps the creating thread's context in a field, and `RequestScopedThreadContextListener.DestroyContext` holds one. I haven't checked all of those, so I can't say yet that it is truly confined to its thread.
- A monitor nobody else is holding costs almost nothing, and the maps are small.

Once we've checked that nothing hands a live `ThreadContext` to another thread, we could replace the `synchronizedMap` with a plain `HashMap` in a follow-up issue. That's a bigger change than this PR should make, though.

container/openejb-core/src/main/java/org/apache/openejb/core/ThreadContext.java
rzo1: Thanks, applied in a728ba18a6.

rzo1:
Full TCK build is ok.

rzo1:
@jungm Just a note: I think this would be backportable to `tomee-10.x`. It only uses the Concurrency 3.0 SPI (`ThreadContextProvider` / `ThreadContextSnapshot`), which 10.x already has. All commits cherry-pick cleanly onto `tomee-10.x`, and `ThreadContextCaptureTest` and `CUTaskFailingContextTest` pass there.

Since it changes when the context is captured (on submit instead of when the task runs), I'd do it as a separate `[tomee-10.x]` PR after this one is merged, and run the TCK against it before merging.

PR #2943, opened 2026-09-17
Additional Fixes for TOMEE-4682 - Do not create session if not needed

exabrial:
Additional Fixes for TOMEE-4682 - Do not create session if not needed

I noticed that the current code always creates a session. I don't believe this is the correct behavior.


rzo1:
Cherry picked on Jons branch directly.

